PR #141, opened 2012-03-27
Candle Chart renderer added, Gallery updated too

jerryzhenleicai:
Please check out the Candler Chart Demo (under Stock Demo). I added a new option called "renderer" which can be either "line" or "candle".


danielkrizian:
Just FYI, the rebirth of the candlestick chart development here: https://github.com/danvk/dygraphs/pull/538

PR #269, opened 2013-09-12
Modifying the numericTicks algorithm to include the last 'nice' tick (either at range end or just past it)

blindmonkey:


danvk:
In light of #506, I'm struggling to remember why this change was made. @blindmonkey do you remember?

blindmonkey:
The reason for this was that we wanted to expand the axis range to the nearest nice ticks and show them for context. So if your data ranges from 1 to 9, we wanted to make sure to show 0 and 10 as ticks as well. This also makes it so that both the start of the range and the end of the range are treated the same.

I'm not sure I understand why #506 is happening. It seems that if we're generating the extreme ticks, they should be more likely to be displayed, not the other way around.

In any case, sorry for taking so long to respond, and it looks the fix won't affect us. Thanks for letting me know!

PR #285, opened 2013-11-16
Make highlight background configurable

foobarth:
Background color and animation while highlighting series
is now configurable using these options and defaults:
    - highlightSeriesBackgroundAnim: true
    - highlightSeriesBackgroundColorRGB: '255, 255, 255'

Updated dygraph-options-reference.js

Needed this for my project and thought it would be a good thing in general.

[Review on Reviewable](https://reviewable.io/reviews/danvk/dygraphs/285)

### DIFF
--- a/dygraph-options-reference.js
+++ b/dygraph-options-reference.js
@@ -160,6 +160,18 @@ Dygraph.OPTIONS_REFERENCE =  // <JSON>
     "type": "float",
     "description": "Fade the background while highlighting series. 1=fully visible background (disable fading), 0=hiddden background (show highlighted series only)."
   },
+  "highlightSeriesBackgroundAnim": {
+    "default": "true",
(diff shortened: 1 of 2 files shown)


danvk:
I'm closing this due to inactivity, but would be happy to re-open if my questions are addressed.

klausw-g:
Sorry about not responding earlier.

I agree that the default white border color will look ugly when using a dark background. I'm not sure if it needs to be separately configurable if it could just use the canvas background color directly, but I didn't find an existing option where this is currently exposed - is this only changeable via CSS on the canvas element itself?

About the anim option - I have run into performance regressions in the past, for example https://code.google.com/p/chromium/issues/detail?id=93753 , but not sure if that is sufficient justification to keep an option around for it. Maybe a global animation toggle would be more appropriate instead of having separate options in each specific case.

PR #311, opened 2014-09-12
Add highlightMinDistance option

panuhorsmalahti:
This new option is required to only highlight a point if the distance is smaller than a specified distance in pixels. The default is Infinity (which doesn't change the existing behaviour in any way). This is useful for e.g. creating scatter plots and adding interaction to clicking the points.

Without this option a point is highlighted even if it's not in clicking distance.

### DIFF
--- a/auto_tests/tests/callback.js
+++ b/auto_tests/tests/callback.js
@@ -529,7 +529,7 @@ CallbackTestCase.prototype.testFailedResponse = function() {
 };
 
 
-// Regression test for http://code.google.com/p/dygraphs/issues/detail?id=355 
+// Regression test for http://code.google.com/p/dygraphs/issues/detail?id=355
 CallbackTestCase.prototype.testHighlightCallbackRow = function() {
   var highlightRow;
   var highlightCallback = function(e, x, pts, row) {
@@ -672,7 +672,7 @@ CallbackTestCase.prototype.testDrawPointCallback_idx = function() {
 CallbackTestCase.prototype.testDrawHighlightPointCallback_idx = function() {
     var idxToCheck = null;
 
-    var drawHighlightPointCallback  = function(g, seriesName, canvasContext, cx, cy, color, pointSizeParam,idx) {
+    var drawHighlightPointCallback  = function(g, seriesName, canvasContext, cx, cy, color, pointSizeParam, idx) {
         idxToCheck = idx;
     };
     var testdata = [[1, 2], [2, 3], [3, NaN], [4, 2], [5, NaN], [6, 3]];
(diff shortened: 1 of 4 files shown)


auto_tests/tests/callback.js
```diff
@@ -692,3 +692,35 @@ CallbackTestCase.prototype.testDrawHighlightPointCallback_idx = function() {
     DygraphOps.dispatchMouseMove(g, 6, 3);
     assertEquals(5,idxToCheck);
 };
+
+/**
+ * Test that highlight min distance works
+  */
+CallbackTestCase.prototype.testhighlightMinDistance= function() {
+    var idxToCheck = null;
```
danvk: please use a two space indent (see http://dygraphs.com/changes.html)

dygraph-options-reference.js
```diff
@@ -148,6 +148,12 @@ Dygraph.OPTIONS_REFERENCE =  // <JSON>
     ],
     "description": "Draw a custom item when a point is highlighted.  Default is a small dot matching the series color. This method should constrain drawing to within pointSize pixels from (cx, cy) Also see <a href='#drawPointCallback'>drawPointCallback</a>"
   },
+  "highlightMinDistance": {
+    "default": "Infinity",
+    "labels": ["Interactive Elements"],
+    "type": "Object",
+    "description": "Minimum distance in pixels to highlight a series."
```
danvk: This description needs to be expanded—for example, it doesn't mention anywhere that this is a Euclidean distance.

dygraph.js
```diff
@@ -1918,8 +1924,8 @@ Dygraph.prototype.findStackedPoint = function(domX, domY) {
     }
     // Stop if the point (domX, py) is above this series' upper edge
     if (setIdx === 0 || py < domY) {
-      closestPoint = p1;
-      closestSeries = setIdx;
+        closestPoint = p1;
```
danvk: two space indent

dygraph.js
```diff
@@ -1812,8 +1813,9 @@ Dygraph.prototype.eventToDomCoords = function(event) {
  * Returns {number} row number.
  * @private
  */
-Dygraph.prototype.findClosestRow = function(domX) {
+Dygraph.prototype.findClosestRow = function(domX, domY) {
```
danvk: This method's docstring no longer reflects what it does. When you have `highlightMinDistance` set, would it make more sense to switch to `findClosestPoint`, rather than `findClosestRow`?

danvk:
I don't have strong opinions about `NaN` values and this option. I actually don't entirely understand what this option is supposed to be doing. Does it prevent selection of a row unless any point in that row is within `highlightMinDistance` pixels?

danvk:
Closing due to inactivity. Feel free to reopen or send a new PR if you'd like to revisit this.

panuhorsmalahti:
The point of this option is to enable clicking of points better. I'd like a point to be highlighted only when it's somewhat close to the point, not when it's far away. The highlight is a cue for the user so that the user knows which point will be clicked. 

I quote my opening message: "Without this option a point is highlighted even if it's not in clicking distance."

PR #314, opened 2014-10-13
New option 'rangeSelectorPlotFillGradientColor' (e.g. for using dygraphs on a dark background)

flesser:
If you use dygraphs on a dark background, a filled range selector does not look very good. The fill gradient always fades to a hard-coded 'white' at the top.
An example can be seen in the attached modified tests/range-selector.html

I added an option to specify a custom color to fade to, so the user can choose the page background color or a transparent rgba() color.


danvk:
Closing due to inactivity, it seems like #548 supersedes this?

PR #469, opened 2014-11-15
Smooth plotter using bezier curves

### LINKED ISSUE #370: Add a "smooth line" plotter
_From [danvdk](https://code.google.com/u/danvdk/) on August 21, 2012 10:15:37_

This is a popular option in other charting libraries and in spreadsheets, e.g. the "function"-style charts in GViz or splines in High Charts ( http://www.highcharts.com/demo/spline-symbols ).

_Original issue: http://code.google.com/p/dygraphs/issues/detail?id=371_

danvk:
Fixes #370, one of our more popular issues.

Here's what it looks like:
screen recording 2014-11-15 at 04 08 pm

This uses cubic splines (aka bezier curves) to smooth out the lines between points. To draw a bezier curve between two points, you must also specify two control points which specify how it bends:

300px-bezier_curve svg

The problem is to determine those control points.

When you connect a bunch of bezier curves, each original point in the series has both a left control point and a right control point. To avoid forming a kink in the chart, the left control point, the series point and the right control point all need to fall on a line.

The algorithm I wound up using follows the HighCharts implementation of this. Here's how it works...

The setup is that you have a point whose left & right controls you're trying to determine. You also know the previous and next point:
setup

Start by placing the control points α of the way along each line segment. This is the only parameter we'll need to control the amount of smoothing:
controls on line

But now we've broken the rule that the original point has to be on the line between the control points! To fix that, we shift the control points up by ∆y, which can be determined with some simple algebra:
controls shifted

But this introduces a new problem: the right control point is above the data points, and so it will create an impression of a false maximum to the right of the data point. To fix this, we cap the y-value of the control points:
controls capped

But now we've re-broken that rule about being on a line. So we have to do the mirror-image adjustment to the left control point:
controls capped and adjusted

and now we have our control points!

Hopefully the algorithm makes good sense now. I implemented this as a separate plotter in `/extras` for now, but it might make more sense to move this into core dygraphs. That would make it easier to have the smooth plots play nicely with other options like `connectSeparatedPoints`, `strokePattern` and `drawPoints`.

### DIFF
--- a/tests/plotters.html
+++ b/tests/plotters.html
@@ -7,7 +7,6 @@
     <script type="text/javascript" src="../excanvas.js"></script>
     <![endif]-->
     <script type="text/javascript" src="../dygraph-dev.js"></script>
-    <script type="text/javascript" src="../extras/smooth-plotter.js"></script>
 
     <script type="text/javascript" src="data.js"></script>
     <style type="text/css">
@@ -17,9 +16,6 @@
       div.chart {
         width: 640px;
         height: 320px;
-      }
-      input[type="range"] {
-        width: 400px;
       }
     </style>
   </head>
@@ -56,9 +52,7 @@
     <div id="mixed-error" class="chart"></div>
 
     <h2>Smooth Lines</h2>
-    <p>This plotter draws smooth lines between points using bezier curves:</p>
-    Smoothing:&nbsp;<input type="range" id="smoothing-amount" min=0 max=0.7 step=0.02 value=0.3><br>
-    <div id="smooth-line" class="chart"></div>
+    <p>See the <a href="smooth-plots.html">smooth-plots demo</a> for an example of a custom plotter which connects points using bezier curves instead of straight lines.</p>
 
     <script type="text/javascript">
       // Darken a color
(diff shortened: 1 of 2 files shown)


Stafie:
@danvk Is there a way to have the `drawPoints` enabled when using the smooth-plotter?

Thank you!

danvk:
Unfortunately, no. Feel free to file an issue (or, better, a pull request!).

One nice solution would be to pull the `drawPoints` behavior out into its own [plotter][1]. That way it would work with any sort of display.

[1]: http://blog.dygraphs.com/2012/08/introducing-custom-plotters.html

bhyatm:
Hi,

The fillGraph option does not seem to be supported when using the smooth plotter - 
Is this supported and if so how do I implement ??

Thanks

shu-frank-csdn:
very good.Thanks for dygraph.

RedShift1:
Would also like to see fillGraph working with the smooth plotter

protyagov:
The **filllGraph** appears working when I specify plotter function per series like so:

```javascript
                // plotter: smoothPlotter,
                series: {
                    "V": {
                        color: "#5997FA",
                    },
                    "H": {
                        plotter: smoothPlotter,
                        color: "orange",
                    }
                },
                fillGraph: true,
```
screen shot 2018-06-12 at 4 24 21 pm

olivier-monaco:
@protyagov fillGraph does not work. Zoom in on your graph and you will see line of the fill, no curves.

I tried to adapt the smoothPlotter to create smoothFillPlotter. See #937

armsp:
@danvk I needed some help with understanding the Bezier plot.  

I understood how you are finding the control points of a given data point. But what I am not able to figure out is -   
The general literature on cubic Bezier curves talks about control points for 2 data points unlike yours. So after you calculate the control point for a single data point, how exactly do you then draw the curves from the previous data point to the next data point and how do you use the control point for drawing the curve ? 
Please do let me know how you did it.

Issue #500, opened 2014-11-29
Remove support for old browsers


danvk:
Which browsers to support? I'm thinking anything that [fully support canvas](http://caniuse.com/#search=canvas), i.e.:
- IE >= 9
- Android >= 4

I'd also like to require [support for ES5](http://kangax.github.io/compat-table/es5/), which largely means the same thing. This gets us things like `Array.prototype.map`, `Array.prototype.forEach` and `Array.isArray`.

If we were willing to drop IE9, we could also get support for [typed arrays](http://caniuse.com/#search=Array), which could be a major performance boost.

What are some things that this would let us hack out?
- [x] excanvas
- [x] event normalization
- [x] All the IE `<meta>` [hacks](http://dygraphs.com/ie.html)
- [ ] Lots of `hasOwnProperty` checks
- [ ] Android hacks to disable clipping areas
- [ ] Lots of `var that = this;` (because of `Function.prototype.bind`)

RedShift1:
IE9 is basically off the map now with 1.07 % market share... Would love to see better support for typed arrays, major gains in memory usage on huge datasets.

PR #504, opened 2014-12-04
Remove excanvas and other IE hacks

danvk:
Well that was fun!

`dygraph-combined.js` shrinks by about 3k as a result of this change.

Related to #500


vivekrajan123:
how we can draw digraph vertically with panning and zooming?
Thanks.

PR #534, opened 2015-01-22
Add cleanup function to the animation selection so that clearing a selec...

cthrax:
...tion always happens.

I have a case where I'm clearing a number of graphs at once and so I am hitting the 30ms time limit sometimes and so my clears were not always happening. This ensures that when you're asking to clear, you actually clear.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -11,7 +11,6 @@
     "compile-with-closure.sh",
     "dashed-canvas.js",
     "data.js",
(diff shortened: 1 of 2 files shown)


danvk:
This looks fine -- could you take a crack at writing a test case for it?

danvk:
Merging. A test would be nice.

PR #541, opened 2015-02-02
Call previously set callbacks and reattach them on detach()

mjanda:


danvk:
Thanks for this. The real solution here is to use event listeners rather than callbacks, but that will have to wait for 2.0.

PR #546, opened 2015-02-11
basic support for multiple y-axes

sweetpi:
Just opening this pull request for discussions.

I started to implement support for multiple y-axes in the axes and labels plugin. It needed some changes in the layout, too. I introduced a "position" setting for axes to select if the axis is shown at the left or right. I use the naming "y3", "y4", .. for the additional axes.

1) The current naming scheme makes it difficult to handle the settings, maybe it would be better to change the axes configuration to an array, but this would break compatibility to the previous versions. What do you think about it?

2) I would like to implement dynamic tick-label widths, so that the width of the labels get adapted to the text length of the containing text at zoom. So the reserved space at left and right could change dynamically. How should this be handled best? My current proposal would be to emit an layout event, if the length of the ticks changes significantly.

There are some tests failing, I will fix them if we agree in the changes.

### DIFF
--- a/dygraph-layout.js
+++ b/dygraph-layout.js
@@ -75,6 +75,10 @@ DygraphLayout.prototype.getPlotArea = function() {
   return this.area_;
 };
 
+DygraphLayout.prototype.getAxisArea = function(axis) {
+  return this.axisArea_[axis];
+};
+
 // Compute the box which the chart should be drawn in. This is the canvas's
 // box, less space needed for axis, chart labels, and other plug-ins.
 // NOTE: This should only be called by Dygraph.predraw_().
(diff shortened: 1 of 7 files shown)


danvk:
dygraphs has never supported more than two y-axes, and we've historically been reluctant to do so. See [this discussion][1] and [this pull request][2]. The gist is that charts with independent 5 y-axes tend to be confusing and misleading in a [Grammar of Graphics][3] sense. And supporting these types of charts adds complexity to dygraphs.

Let me know what you think.

[1]: https://groups.google.com/forum/#!searchin/dygraphs-users/three$20y-axes/dygraphs-users/T8fBrEsV0_w/2RjS6DpSRlMJ
[2]: https://github.com/danvk/dygraphs/pull/53
[3]: http://www.amazon.com/The-Grammar-Graphics-Statistics-Computing/dp/0387245448

Issue #532, opened 2015-01-12
feature request: add option to configure style of range selector line


mg1075:
I really like the range slider, 
except it would be nice to be able to style the range slider's interactive layer outline

![image](https://cloud.githubusercontent.com/assets/2031435/5233374/92292232-7742-11e4-80d0-4213c968425f.png)

Currently, the line is hard-coded to a width of 1.5, 
and this makes the range selector line stand out too much from the real data viz. 

In addition to providing an option for styling, 
it might also be worth considering a change to the slider's default outline width 
from 1.5 to a lower value.

https://github.com/danvk/dygraphs/blob/b94ba4b52e227816448775cce9687f33a80cfe9a/plugins/range-selector.js#L619

Thanks.

PR #548, opened 2015-02-14
A few basic range selector options

blcook223:
I added the following options to the range selector plugin: 

- rangeSelectorBGStrokeColor
- rangeSelectorBGLineWidth
- rangeSelectorPlotLineWidth
- rangeSelectorFGStrokeColor
- rangeSelectorFGLineWidth
- rangeSelectorAlpha

I also added these options to the whitelist of options that do not change pixel position in `dygraph-utils.js`, documented the options in `dygraph-options-reference.js`, and added an auto-test. These changes address open issue #532. I left the range selector gradient color as a hard-coded `'white'` to avoid possible conflicts with PR #314.

Incidentally, it looks like my editor deleted some whitespace in some of the files I modified.

[Review on Reviewable](https://reviewable.io/reviews/danvk/dygraphs/548)


danvk:
@paulfelix thoughts? We need to get some kind of story in place for plugin-specific options.

paulfelix:
Looks great. Might be nice to have the #314 feature folded into this request somehow, maybe with a new option name consistent with Benjamin's names.

danvk:
This seems reasonable, my only concern is that this will make it harder to migrate range selector options to a separate dictionary later on if we want to do that, i.e.

```javascript
{
  rangeSelector: {
    height: 40,
    plotStrokeColor: "#808FAB",
    plotFillColor: "#A7B1C4",
    backgroundStrokeColor: "gray",
    backgroundLineWidth: 1,
    plotLineWidth:1.5,
    foregroundStrokeColor: "black",
    foregroundLineWidth: 1,
    alpha: 0.6
  }
}
```

vs.

```javascript
{
  rangeSelectorHeight: 40,
  rangeSelectorPlotStrokeColor: "#808FAB",
  rangeSelectorPlotFillColor: "#A7B1C4",
  rangeSelectorBackgroundStrokeColor: "gray",
  rangeSelectorBackgroundLineWidth: 1,
  rangeSelectorPlotLineWidth:1.5,
  rangeSelectorForegroundStrokeColor: "black",
  rangeSelectorForegroundLineWidth: 1,
  rangeSelectorAlpha: 0.6
}
```

Another option would be to find a way to pull this in via CSS. Styling a `<canvas>` via CSS isn't usual (this is one of the big drawbacks of using canvas vs SVG), but we might be able to force it by creating a temporary DOM element inside the chart w/ a known class name & getting its color, strokeWidth, alpha, etc. via `getComputedStyle`. It's a hack & I've never seen anyone do it this way before.

@paulfelix @klausw-g @kberg thoughts?

Specific feedback:
- You may as well incorporate #314 into this—I've closed it out.
- Let's write out `Foreground` and `Background` rather than `FG` and `BG`.
- Please modify the "custom range selector height and plot color " chart in `tests/range-selector.html` to show off these new options.
- It might be helpful to draw an "anatomy of the range selector" chart to keep all this straight.

blcook223:
Thanks for taking a look at this. I've made some of the specific requested changes. I'll keep working on this and make the rest of the changes probably some time this week.

klausw-g:
I don't have a strong preference here. The nested option dictionary looks a bit cleaner, but a downside is that it's less friendly to automated searches - "rangeSelectorHeight" is easy to find in source or documentation, but finding "height" in the context of a rangeSelector option object (possibly passed in via a differently-named reference) can get tricky.

The CSS option does seem unusual - is it safe to assume that custom properties will get propagated consistently if the browser doesn't understand them? There's "CSS Custom Properties For Cascading Variables" but I'm unsure how supported that is: http://thatemil.com/blog/2013/12/15/playing-around-with-css-variables-custom-properties/ . My gut feeling is that this is likely to be too tricky and explicit inheritance via code would be clearer, especially if the API has helper methods to make this easy to manage.

blcook223:
I added some images to try to show the "anatomy of the range selector" to tests/range-selector.html. Not sure if this is what you had in mind. Let me know if you have other ideas.

danvk:
Thanks! I'll merge. The images aren't quite what I had in mind, but I think I can take it from here.

danvk:
See #564, particularly [this image](https://github.com/danvk/dygraphs/blob/range-selector-example/tests/range-selector-anatomy.png).

PR #550, opened 2015-02-18
legend: follow positioning fixes

mjanda:
I've run into issue with legend position when it was clipped by overflow: hidden (using fixed width modal to show the chart), so I think it might make more sense to switch position when it would go outside chart area.

Also proposing 20->50px to show more points around currently highlighted one. When you move your mouse quickly, it might jump inside legend area which then stops following and stays put - especially when you have wider gaps in your data (even as you move your mouse over the gap, legend stays in one place until you reach next point).


danvk:
Thanks for the tweaks!

PR #557, opened 2015-03-06
Add getRowForX method & fix synchronizer extra

### LINKED ISSUE #554: Synchronization extension skips on other charts
On the example on the website, the synchronization misses intervals as you move around. You'll notice as you slide around, the chart you're hovering over updates each interval. But the other charts skip intervals. You can see the problem here:

http://dygraphs.com/gallery/#g/synchronize

Notes:
* Seems to consistently skip the same intervals between page reloads
* Easy to see if you hover over the highest peak in the chart
* Same behavior in Chrome and Firefox on Windows.
* For my charts (like the example) I'm using very regular data-- the same exact intervals for each chart. They are each one day so it's not like the charts are off by hours or minutes or seconds.

danvk:
Fixes #554 

The bug was of the one character sort (d5ce61b). This has come up a few times, so I moved the binary search into a new public dygraphs method, `getRowForX`, gave it well-defined behavior and tested it.

[Review on Reviewable](https://reviewable.io/reviews/danvk/dygraphs/557)


danvk:
@userlite FYI
@kberg mind taking a look?

danvk:
@kberg i'll merge this EOD if you don't get a chance to look.

PR #559, opened 2015-03-13
Allow setVisibility() to set the visibility of multiple series at once

sslepian:
Currently, each visibility change triggers a redraw, which causes functionality like hiding/showing all series (or applying a filter) to be very slow. This change allows setVisibility() to accept an array instead of a single numbers so that the redraw only happens after the visibility of multiple series has been updated.


sslepian:
The first commit doesn't actually work because I somehow failed to push half of it - the new one contains the other half of the change.

danvk:
Thanks for the change! I'm surprised there wasn't already an API for this.

Could you add a test case? (in auto_tests)

sslepian:
I'll look at it, there wasn't one already for it in auto_tests (just one in "tests" which is are manual tests as far as I can see) so I need to figure out how to make one. I can't test if the lines are rendered I think since it's canvas and has no dom elements, so I guess I'd test by just calling visibility() before and after? If you have a better suggestion please mention it and I'll get it done.

danvk:
That would be a reasonable way to do it. You can also call `setSelection` and inspect the legend, which is a DOM element & should only reflect visible series. There's even a [helper][1] to get its contents. The test should make sure that both the old `setVisibility` and the new one work.

[1]: https://github.com/danvk/dygraphs/blob/dc836ebb77ab18a402ea4ae590059299d88b0deb/auto_tests/tests/Util.js#L54

sslepian:
I closed out this pull request because I realized that I need it on its own branch in my repo (still new to how github handles pull requests). Updated request coming shortly.

PR #560, opened 2015-03-16
update setVisibility to be able to set the visibility of multiple series at once

sslepian:
Currently, each visibility change triggers a redraw, which causes functionality like hiding/showing all series (or applying a filter) to be very slow. This change allows setVisibility() to accept an array instead of a single numbers so that the redraw only happens after the visibility of multiple series has been updated.

This replaces https://github.com/danvk/dygraphs/pull/559

### DIFF
--- a/dygraph.js
+++ b/dygraph.js
@@ -3602,19 +3602,25 @@ Dygraph.prototype.visibility = function() {
 };
 
 /**
- * Changes the visiblity of a series.
+ * Changes the visibility of one or more series.
  *
- * @param {number} num the series index
+ * @param {number|number[]} num the series index or an array of series indices
  * @param {boolean} value true or false, identifying the visibility.
  */
 Dygraph.prototype.setVisibility = function(num, value) {
   var x = this.visibility();
-  if (num < 0 || num >= x.length) {
-    console.warn("invalid series number in setVisibility: " + num);
-  } else {
-    x[num] = value;
-    this.predraw_();
+
+  if (num.constructor !== Array) num = [num];
+
+  for (var i = 0; i < num.length; i++) {
+    if (num[i] < 0 || num[i] >= x.length) {
+      console.warn("invalid series number in setVisibility: " + num[i]);
+    } else {
+      x[num[i]] = value;
+    }
   }
+
+  this.predraw_();
 };
 
 /**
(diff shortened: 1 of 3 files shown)


danvk:
Looks good, thanks for the test!

danvk:
Actually, I'm wondering if this is the right API. Say you want to change the visibility of a few series. With this API, you have to split the series into those which should be visible and those which shouldn't, and then make two calls to `setVisibility`.

I wonder if a better API would be for:
1. [`visibility`][1] to clone the visibility array (rather than returning it directly) 
2. `setVisibility` to take an array of booleans

That way you could get the visibility array, modify it as you want and then do the update all in one fell swoop.

[1]: http://dygraphs.com/jsdoc/symbols/Dygraph.html#visibility

sslepian:
I thought about doing something similar, and decided to go with the current solution because it reduces the needed calls to a maximum of two without breaking API compatibility - if someone was using setVisibility() with a single series before, it'll continue just working.

Xan-Kun:
I totally agree with danvk. It fact, this was the behaviour (getting a copy of the array, manipulating it and feeding it back to the setVisibility function) I expected and it didn't work first, so I had to see in the source and understand, that I really had to set every single in separately.

Another approach would be to get the reference to the array, manipulate it and manually call somthing like redraw,, but that seems extremely invasive to me.

I have this exact problem in a page with 3 graphs and 16 series per graph. They are all synchonized and the page behaves extremely responsive, except when setting the visiblity of all except one. It takes over a second due to 3 * 16 redraws.

dygraph.js
```diff
@@ -3602,19 +3602,25 @@ Dygraph.prototype.visibility = function() {
...
  */
 Dygraph.prototype.setVisibility = function(num, value) {
   var x = this.visibility();
-  if (num < 0 || num >= x.length) {
-    console.warn("invalid series number in setVisibility: " + num);
-  } else {
-    x[num] = value;
-    this.predraw_();
+
+  if (num.constructor !== Array) num = [num];
```
reinert: This LOC is blocking me to use this function. The constructor check, although fast, is not reliable. It can fall into certain corner cases which will fail to determine the variable as an array (false negative). E.g., when your array is from different window contexts. The foolproof solution is to use `Array.isArray()` method. See http://web.mit.edu/jwalden/www/isArray.html for more info.

As I'm using Dygraphs inside a GWT app, I'm falling into this corner case.

danvk: FYI I merged your PR #648. Thanks!

reinert: Thanks you for this awesome library. I'm dealing with millions of records in multiple graphs dynamically and only dygraphs could handle it gracefully. (I tested four libraries before this one)! Thank you!

danvk: Cool! If you've got something public to add to `users.html`, you know how to send a PR :)

PR #565, opened 2015-03-21
Migrate to gulp & karma

danvk:
This is heavily based on (and supersedes) #528.

Highlights:
- Moves all code into `src/`
- Switches from jstd to [Mocha][], [Chai][] and [Karma][].
- Uses [gulp][] for building

The upshot is that dygraph becomes a more modern JavaScript library and we can start using nice tools like Istanbul (for code coverage) and Babel (for ES6/7→ES5 transpilation).

[mocha]: http://mochajs.org/
[chai]: http://chaijs.com/api/assert/
[karma]: https://karma-runner.github.io/0.12/index.html
[gulp]: http://gulpjs.com/


danvk:
@kberg @klausw-g @flooey FYI

This is a fairly big change to dygraphs—see PR description. I'd like to merge this in relatively soon (tonight or tomorrow), so please let me know if you have any objections.

flooey:
It looks like this only changes the build process, not any of the code, right?  If that's the case, it'll be a bit annoying to set up the next time we update our copy, but should be fine otherwise.  Feel free to go ahead.

danvk:
That's correct. To get the same file as before, you run:

```
npm install
gulp dist
ls dist/dygraph-combined.js
```

dygraphs already depends on NPM, so this shouldn't be a big deal.

If anyone depends on individual JS files in dygraphs, they'll have to be updated. But I'd highly recommend not doing that. I plan to start using ES6 features soon, which will make these files not directly includable in the browser.

danvk:
Okie doke, this is going in.

PR #576, opened 2015-04-05
Fix range selection when chart is located inside fullscreen element

mjanda:
Hello,
when you put chart inside element that's fullscreen (using fullscreen api), range selections start at wrong positions (current findPos clearly uses original position, not zoomed).

If I'm not mistaken, original findPos uses some 12 years old code that was created because getBoundingClientRect didn't exist back then. So I'm proposing this simpler version. Or is there some reason to keep the old one?

### DIFF
--- a/src/dygraph-utils.js
+++ b/src/dygraph-utils.js
@@ -179,8 +179,8 @@ Dygraph.findPos = function(obj) {
       d = document.documentElement;
 
   return {
-    x: p.left + (w.pageXOffset || d.scrollLeft) - (d.clientLeft || 0),
-    y: p.top  + (w.pageYOffset || d.scrollTop)  - (d.clientTop  || 0)
+    x: p.left + (w.pageXOffset || d.scrollLeft),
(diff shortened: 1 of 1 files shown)


danvk:
I like it :)

Since this is a slightly frightening change, could you check that the Karma/mocha tests still pass under a few non-PhantomJS browsers, e.g. Firefox, Safari & IE9?

mjanda:
Firefox (37) and Safari (8 on Yosemite) both pass, IE9 and IE10 fail many tests both before and after change, so hard to tell if this change broke something new.

danvk:
I'm slightly concerned by [this question][1] on Stack Overflow, but I'll go ahead and merge. If there are cross-browser issues, we can deal with them during QA for the 2.0.0 release.

Thanks for bringing `findPos` into the modern age :)

[1]: http://stackoverflow.com/questions/10230715/why-getboundingclientrect-gives-different-values-in-ie-and-firefox

PR #579, opened 2015-04-10
Add panToTime method

adamborowski:
I've added one method that is handy if you want to highlight a point that might be not visible at the moment, because any zoom is applied.
It does panning left or right. This is useful when you are doing something like table-chart synchronization.

### DIFF
--- a/src/dygraph.js
+++ b/src/dygraph.js
@@ -3607,6 +3607,32 @@ Dygraph.prototype.size = function() {
   return { width: this.width_, height: this.height_ };
 };
 
+
+/**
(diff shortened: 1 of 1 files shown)


danvk:
Thanks for this change! I could see it being useful. A few comments:

- This needs a unit test
- Please try to follow the style of the [dygraphs coding style][1], e.g. two space indents.
- `panToTime` is a bit too specific, since the x-axis could also be numeric.
- What does this do if the given time is already visible? Does it center the view? It's worth documenting.

[1]: http://dygraphs.com/changes.html

danvk:
Closing this out for now, but feel free to re-open.

PR #583, opened 2015-04-16
Dygraph.findPos may return NaN

clelom:
I embedded a Dygraph into a webcomponent showing several graphs within a elements generated by <template repeat>. Clicks within the graphs did not work correctly. The reason seems to be that the template created a document-fragment around the dygraph element which returned NaN on obj.scrollLeft and obj.scrollTop. The while loop going through all parent nodes up to document.body would pass this document-frament and carry on the NaN returned there.

The consequence was that the x and y returned from Dygraph.findPos resulted in NaN and subsequent calculations failed.

### DIFF
--- a/src/dygraph-utils.js
+++ b/src/dygraph-utils.js
@@ -201,8 +201,8 @@ Dygraph.findPos = function(obj) {
 
   // This handles the case where the object is inside a scrolled div.
   while (obj && obj != document.body) {
-    curleft -= isNaN(obj.scrollLeft) ? 0 : obj.scrollLeft;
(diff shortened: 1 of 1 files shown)


danvk:
Thanks for the change! Could you add a unit test for this? This is a pretty obscure scenario which seems unlikely to be tested in the future otherwise.

danvk:
Would this be fixed by #576?

danvk:
Closing in favor of #576; please let me know if that doesn't fix your issue.

clelom:
Thanks, and yes, #576 looks promising, but I will only get around to test it in a few days.

PR #585, opened 2015-04-19
allow custom bar series to have the bar continue across missing values

jmarshallnz:
This is really a convenient hack more than anything else.  Many graphs I produce tend to have a data series containing missing values, along with a model fit (and standard errors) which are continuous across the missing values in the data. You only want the data showing up in legends and the like, thus combining the multiple series by using the custombars dataHandler.

I could produce another dataHandler for this, but it would be identical to bars-custom.js with the exception of this change, which seems overkill.  If there's a smarter way to do this while still keeping the behaviour of inherited missing-ness in the bars as is present currently, am happy to implement (am a javascript noob).

### DIFF
--- a/src/datahandler/bars-custom.js
+++ b/src/datahandler/bars-custom.js
@@ -43,11 +43,7 @@ CustomBarsHandler.prototype.extractSeries = function(rawData, i, options) {
     // Extract to the unified data format.
     if (point !== null) {
       y = point[1];
-      if (y !== null && !isNaN(y)) {
-        series.push([ x, y, [ point[0], point[2] ] ]);
-      } else {
-        series.push([ x, y, [ y, y ] ]);
-      }
+      series.push([ x, y, [ point[0], point[2] ] ]);
     } else {
       series.push([ x, null, [ null, null ] ]);
     }


danvk:
This is definitely a behavior change. It can't be made so simply.

I'm not sure I entirely understand your use case. Adding a demo of it in `tests/` would help clarify what you're trying to do.

If this is something we really do need, the right way to enable it would be via an option, either a new or existing one. There's a [`connectSeparatedPoints`][1] option and also [`stackedGraphNaNFill`][2].

You could also split out your bars into an independent series and use a [plotter][3].

[1]: http://dygraphs.com/options.html#connectSeparatedPoints
[2]: http://dygraphs.com/options.html#stackedGraphNaNFill
[3]: http://dygraphs.com/tests/plotters.html#mixed-error

jmarshallnz:
Agreed. I ended up writing a custom dataHandler and plotter.  The dataHandler was largely copy and paste, but isn't too bad as I took the opportunity of adding an additional y value for plotting (data, fit, lwr, upr).

One thing that may be nice is to be able to specify a dataHandler per series. At the moment there's global attributes (customBars, errorBars, and dataHandler) for handling when to use what, and one needs to set one of custombars/errorBars to true if you wish to call the errorPlotter from a custom plotter. I think this would go away somewhat if these were per-series switches (similar to how plotter is specified).

Would this be something you'd be interested in?

danvk:
Yes, definitely! I'll go ahead and close this PR, but I'd be very much interested in per-series data handlers.

PR #586, opened 2015-04-20
Load dygraph-combined when loading from NPM

aldendaniels:
Today, when loading dygraphs from NPM via browserify you need to include:

```javascript
var Dygraph  = require('dygraphs/dygraph-combined.js');
```

Being able to do this:
```javascript
var Dygraph  = require('dygraphs');
```

Is preferable.

**UPDATE**
Dygraphs doesn't have a `module.exports` statement, so it can't be loaded via Browserify at all. This PR also includes a fix for this.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "dygraphs",
   "version": "1.1.0",
   "description": "dygraphs is a fast, flexible open source JavaScript charting library.",
-  "main": "dygraph.js",
+  "main": "dygraph-combined.js",
   "directories": {
     "doc": "docs",
     "test": "tests"
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "dygraphs",
   "version": "1.1.0",
   "description": "dygraphs is a fast, flexible open source JavaScript charting library.",
-  "main": "dygraph-combined.js",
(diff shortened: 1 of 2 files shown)


danvk:
Thanks for this change!

I see you switched from `dygraph-combined.js` to `dygraph-combined-dev.js`. The latter includes more assertions and checks for correctness but may be slower. Any particular reason for the switch? The dev version seems like a reasonable default, but I'm curious what your thinking was here and if there are any standard practices for NPM.

aldendaniels:
Hey @danvk - I think referencing the minified version is an anti-pattern for node/NPM as this makes debugging more challenging. The ideal solution would probably be to reference a non-minified version of the production code. 

It also looks like the package is missing the `modules.exports` statement, so I added a commit for that.

Thanks for all the work you've put into Dygraphs BTW - I've really enjoyed using it!

aldendaniels:
Would be really great to backport these changes to a `v1.1.1` and deploy to NPM, since the current NPM version is basically useless w/out the `module.exports` statement.

danvk:
Thanks for the fix. I filed #587 to put this in 1.1.1, which I really should get around to releasing.

aldendaniels:
@danvk - Thanks!

PR #589, opened 2015-04-21
Added HowManyDiapers.com to users.html page

stephenjbell:
Dan, in our email conversation the other day, you suggested that you would consider adding us (HowManyDiapers.com) to the user list if I made a pull request. I've tried to match the formatting of the other items on the page as closely as possible. Thank you for your consideration.


danvk:
Beautiful, thanks!

PR #602, opened 2015-04-22
Add an option for making the highlight background configurable

sslepian:
This makes it possible to configure the highlight background - loosely based on the discussion in https://github.com/danvk/dygraphs/pull/285 (though not using any code in the pull request).

Tests are a bit awkward because the highlight fades in over time, thus each test needs a ~1s delay which slows down the test suite. If there's a better way to handle the delay, I'm all for it.


danvk:
Sorry for the very slow response. I'm happy to merge this change, though I would like to get the tests to run faster.

The easiest way to do that is probably to replace `Dygraph.repeatAndCleanup` with a version that immediately calls the last frame.

danvk:
Closing in favor of #682

PR #615, opened 2015-05-27
Allow fillAlpha to be set on a per-Series basis when fillGraph option is used.

t-lynch:
Changed the way that the fillAlpha option is processes in the _fillPlotter function in order to allow for per-series configuration.   Previously when fillGraph was set, only the global option was recognized.

### DIFF
--- a/auto_tests/tests/per_series.js
+++ b/auto_tests/tests/per_series.js
@@ -55,6 +55,49 @@ it('testPerSeriesFill', function() {
   assert.deepEqual([255,0,0,38], sampler.colorAtCoordinate(6.5, 0.5));
 });
 
+it('testPerSeriesAlpha', function() {
+  var opts = {
+    width: 480,
+    height: 320,
+    axes : {
+      x : {
+        drawGrid: false,
+        drawAxis: false,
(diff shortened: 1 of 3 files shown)


danvk:
Thanks! Mind adding a quick unit test for this?

t-lynch:
Sure.   It may take a day or so before I can get to it.    I did do extensive testing on it in the project I'm using Dygraphs for, and once I push those into production on my end I'll revisit the unit test.

danvk:
Sounds good. We'll wait for a tested PR.

t-lynch:
Ok, that didn't take nearly as long as I thought it might.   Thankfully the instructions were very clear and concise.

Hopefully I took the right approach here - I added a new test to show the different ways the fillAlpha value might be populated,
* unspecified (use default value)
* globally
* per series for all series
* mix of global and per series values

I also amended the existing auto_test for the per-series options and added a new test which ensures that the per-series values are in fact used when filling the series.

Let me know if there is anything else you need to verify this PR.

danvk:
Merged, thanks for adding the unit test!

PR #634, opened 2015-06-24
Fix previous callbacks

mjanda:
My previous solution to keeping/re-attaching callbacks on detach() was clearly completely wrong, so this should hopefully fix it


danvk:
@mjanda thanks for fixing this! @kanitw thanks for reviewing.

Since this was broken before, could you add a regression test for it?

mjanda:
Ok, I've tried

danvk:
Looks great, thanks for the test!

PR #645, opened 2015-07-16
Fix XY series highlight positioning

121watts:
### What
Refactored `findClosetPoint` so that if there is a tie on the X-Axis, the tie-breaker is the point with the closest Y-Axis.  

### Why
Before this commit, when mousing into a graph with highlightSeriesOpts set, the highlighted point / series is the shortest distance from the cursor.  This sounds correct but produces results like this:

screen shot 2015-07-15 at 9 39 24 pm

With this change, the correct points and series are highlighted:

screen shot 2015-07-15 at 9 39 00 pm

### DIFF
--- a/src/dygraph.js
+++ b/src/dygraph.js
@@ -1833,21 +1833,34 @@ Dygraph.prototype.findClosestRow = function(domX) {
  * @private
  */
 Dygraph.prototype.findClosestPoint = function(domX, domY) {
-  var minDist = Infinity;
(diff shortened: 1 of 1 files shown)


danvk:
Sorry for the slow response. This PR doesn't make much sense to me. Having a tie for the x-distance is a very brittle thing. Couldn't you move your cursor over by epsilon in the screenshot and have it be "broken" even after your change?

When you draw a vertical bar, it seems like what you really want is to select the closest point by x-distance alone.

Issue #407, opened 2014-10-20
Add a legendFormatter option


danvk:
_From [roche....@gmail.com](https://code.google.com/u/100735624287815865174/) on August 14, 2013 04:34:20_

Hi,

First, thank you for your great framework ! I use it to display massive data (1 000 000 points) and it works very well !

I would like to suggest a little enhancement: to have a legend callback. So, we can personalize the display and if we don't want some informations on specific lines, we can hide these !

Example:
new Dygraph(..., [], {
   ...
   legendCallback: function(label, color, dash) {
      // Base on generateLegendHTML function
      return "<span style='font-weight: bold; color: " + color + ";'>" +
          dash + " " + label + "</span>";
   }
}

Best regards

Julien Roche

_Original issue: http://code.google.com/p/dygraphs/issues/detail?id=485_

PR #646, opened 2015-07-24
Added labelsSortByValue option so that the legend will display series…

chadwickh:
… by y Value, in decreasing order


danvk:
This feels overly specific. What would really be helpful is some way to plug in your own labels --> HTML converter.

chadwickh:
Would an option legendFormatter that takes a function and returns the formatted HTML make sense?

danvk:
Yes, that would be more reasonable.

danvk:
This is #407

PR #657, opened 2015-09-15
Fixing issue with xRangePad being ignored on unzoom

kbaggott:


danvk:
Thanks for catching & fixing this!

PR #674, opened 2015-10-23
Use ES6 modules

### LINKED ISSUE #376: Remove dygraphs-combined.js from source control
_From [konigsb...@google.com](https://code.google.com/u/101959970229003594722/) on November 09, 2012 09:02:02_

https://github.com/danvk/dygraphs/blob/master/dygraph-combined.js says

This is not the file you are looking for.
A reasonably up-to-date version can be found at http://dygraphs.com/dygraph-combined.js dygraph-combined.js is a "packed" version of the larger dygraphs JS files. It is
smaller and loads more quickly, but is harder to debug.

To generate this file, run "make" or generate-combined.sh.

It should say:

alert("You don't have a proper version of dygraphs-combined. Nothing will work. Nothing at all. Run 'make' or 'generate-combined.sh'. See http://blahblahblah for details")

This will make n00bs lives easier.

_Original issue: http://code.google.com/p/dygraphs/issues/detail?id=393_

### LINKED ISSUE #453: Use a module system
It's a bit ridiculous that we manage all these JS files [manually](https://github.com/danvk/dygraphs/blob/master/dygraph-dev.js). [browserify](http://browserify.org/) is fairly standard these days and solves the problem by adding a node.js-like `require` statement to JS.

One problem is that the various classes that Dygraphs defines don't map cleanly to individual files. For example, at least six or seven different JS files define methods on the `Dygraph` object.

- [x] Update developer documentation (`DEVELOP.md`)
- [x] Make sure copyright stays on `dygraph-combined.js`
- [x] Make sure `tests/*.html` doesn't break
- [x] Set module attributes on `Dygraph` for backwards compatibility, e.g. `Dygraph.dateTicker`
- [x] Update gallery
- [x] Get code coverage working
- [x] Remove gulpfile (& port gwt task over)
- [x] Set the `DEBUG` variable via `envify`
- [x] Get the `DygraphOptionsReference` checks working again
- [x] Migrate the remaining data handlers
- [x] Update tests
- [x] Update `dygraph-gviz.js`
- [x] Kill `dygraph-dev.js`

danvk:
This PR migrates dygraphs from its own ad-hoc dependency system to the new ES6 module system.

Highlights:
- We use [babel][] to transpile ES6 code to the ES5 code that browsers like
- We use [browserify][] to build bundled JS for the browser. Since browserify isolates each module for us, there's no more need for [IIFE wrappers][iife].
- Each file is now its own distinct module, with its dependencies listed explicitly. No more defining properties on `Dygraph` from a dozen different files.
- We use [envify][] to set the `DEBUG` property to distinguish dev/prod builds.
- Tests can be run & iterated on using the mocha HTML test runner and a new `watch.sh` script.
- `DEFAULT_ATTRS` gets moved into its own module.
- Deleted `dygraph-combined.js` from source control (fixes #376)
- A few functions (e.g. `numberAxisLabelFormatter`) move out of `dygraph.js` and into `dygraph-utils.js`. We need to shrink `dygraph.js`, it's huge.

Fixes #453 

[babel]: https://babeljs.io/
[browserify]: http://browserify.org/
[envify]: https://github.com/hughsk/envify
[iife]: http://benalman.com/news/2010/11/immediately-invoked-function-expression/


danvk:
(code coverage doesn't look [quite right][1])

[1]: https://coveralls.io/builds/3936239/source?filename=src%2Fdygraph-utils.js

timeu:
Grest to see that dygraphs is moving forward.
I am wondering if it still will be possible to generate a dygraphs-combined.js to include for example for a GWT wrapper?

danvk:
Running `scripts/build.sh` from `master` should produce `dist/dygraph.min.js` and `dist/dygraph-combined.js` (which are the same thing—I'd like to move to the `.min.js` naming convention for 2.0). It'll also make `dygraph-gwt.jar`.

PR #682, opened 2015-10-27
Add highlightSeriesBackgroundColor option

danvk:
This revives #602.
See also #285 

This is a WIP — trying to speed up the selection fade animation exposed an assumption that `animateSelection_` is making about how `repeatAndCleanup` works. It assumes that if it asks for 10 frames, it will get 10 frames. But that's not necessarily the case.


danvk:
I'm going to merge this in since the option works and is useful. The only issue is that the test is a bit slow. This is annoying and noticeable enough that I suspect I'll fix it if left to my own devices.

Issue #685, opened 2015-11-06
xAxis zooming is broken when using dateWindow of Date


ctron:
Setting a `dateWindow` in the options of type `Date`, will break the zoom X axis functionality.

danvk:
This behavior is exactly what you'd expect http://dygraphs.com/options.html#dateWindow

> Initially zoom in on a section of the graph. Is of the form [earliest, latest], where earliest/latest are milliseconds since epoch. If the data for the x-axis is numeric, the values in dateWindow must also be numbers.

ctron:
And if the data for the x-axis is not numeric?

danvk:
> Is of the form [earliest, latest], where earliest/latest are milliseconds since epoch.

PR #686, opened 2015-11-06
fixes danvk/dygraphs#685: Convert Dates to milliseconds if necessary

ctron:
The xAxisRange method may return Date or Number (milliseconds), this
will cause the doZoomX function to fail since it calls toDataXCoord
which gets a string instead, containing the Date.toString part and
the actual milliseconds offset.

This patch convert Date instances for xRange array into numbers.

### DIFF
--- a/src/dygraph.js
+++ b/src/dygraph.js
@@ -627,7 +627,12 @@ Dygraph.prototype.toDataXCoord = function(x) {
   }
 
   var area = this.plotter_.area;
-  var xRange = this.xAxisRange();
+  var xRangeAny = this.xAxisRange();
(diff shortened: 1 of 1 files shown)


danvk:
I'm open to improving this, but the way to do it would be either by:

1. Converting `Date` objects to millis in `updateOptions`
1. Warning about passing a `Date` object in via `dateWindow`.

ctron:
So I would vote for #1, since #2 still breaks the chart.

danvk:
I'd accept a well-tested PR that implemented the first option.

danvk:
(Closing due to inactivity. Feel free to reopen if you want to continue this.)

Issue #691, opened 2015-11-16
Drawing selection points when pt.canvasy === 0


musicist288:
I came across a behavior that I thought was strange and am wondering what the intent is behind it. 

On this line: https://github.com/danvk/dygraphs/blob/master/src/dygraph.js#L1818 the check to `utils.isOK(pt.canvasy)` seems strange given that isOK includes `0` as a false value. I understand checking for a `NaN` or other falsey values, but why blacklist `0` as well? I thought this was a bug except it also occurs in the legend plugin here: https://github.com/danvk/dygraphs/blob/master/src/plugins/legend.js#L271

Why should values that would draw at 0 in canvas coordinates be excluded both from drawing and from showing up in the legend? Is this behavior intended?

danvk:
Probably not. Can you try changing it to a `NaN` / `null` check and see if any of the tests fail?

musicist288:
The tests pass if I change the checks as in the attached patch file or if I change utils.isOK to:

```JS
export function isOK(x) {
  return x === 0 || (!!x && !isNaN(x));
};
```


[0001-Bug-Allow-points-where-canvasy-0.patch.txt](https://github.com/danvk/dygraphs/files/35861/0001-Bug-Allow-points-where-canvasy-0.patch.txt)

danvk:
Please don't change `isOK`. I'd be happy to accept a PR that changes the two uses you mentioned (`updateSelection_` and the legend), especially if it includes a test where it makes a difference in behavior.

PR #692, opened 2015-11-16
Allow selected points where canvas-y coordinate is 0

musicist288:
As discussed in https://github.com/danvk/dygraphs/issues/691

The check for `utils.isOK` blacklists the value 0. In the context of
validating selected points and reporting them in the legend, any point
where canvas-y zero should be included.


musicist288:
Ok. Updated and tests pass.

danvk:
Perfect, thanks! We'll see if anyone complains.

Issue #371, opened 2014-10-20
Support "select closest point" in addition to "select closest x-value"


danvk:
_From [k.nielse...@gmail.com](https://code.google.com/u/115719069660207256565/) on August 24, 2012 04:32:08_

Hallo dygraph devs

We use dygraph to display relatively dense data plots with multiple series and these never has common x-values. Having the data read out in dygraph is a really nice feature, but it does not work so well under these conditions, because the read out will be constantly switching between the different series and never show all of them. I have illustrated the problem with this fiddle example: http://jsfiddle.net/eM2Mg/272/ It seems that the way it works now is that dygraph locates the x-value closest to the mouse and then displays "all" the y-values for that x. It would be nice if there was an option for the data read out so that it would instead locate a number of nearest x's (one for each series) and then display those x-values and their respective y-values.

instead of only showing the point of the x nearest to the mouse pointer to instead showing one point per series.

Is should be mentioned that the rolling average to some extend solve it, but we would rather not be looking at averaged data.

Just for completeness this is an issue in both google chrome and firefox on linux.

I hope you can help.
Regards Kenneth

_Original issue: http://code.google.com/p/dygraphs/issues/detail?id=374_

musicist288:
Just came across this and want to second this request.  Glad to see it in the 2.0.0 milestone.

musicist288:
Is there a spec in mind for this? I added a shim into a custom build to get the behavior I need but I wouldn't be opposed to implementing a general solution for this request.

danvk:
I'd imagined something like:

```javascript
{ selectMode: 'closest-x' }
```

vs.

```javascript
{ selectMode: 'euclidean' }
```

danvk:
Alternatively, you could imagine making this a callback and having named functions for presets:

```javascript
{ selectMode: Dygraph.findClosestByX }
```

```javascript
{ selectMode: Dygraph.findClosestByEuclidean }
```

I'd be fine with either of those APIs. The latter has the advantage of being more general and catching typos in the JS engine, rather than in dygraphs.

danvk:
Also: I'd be very happy to see a PR for this!

squaregoldfish:
I'd like to resurrect this issue and work on it, but given there's been a couple of partial attempts at it I'm not sure where to begin - take one of those as a starting point or just start from scratch. Any suggestions?

danvk:
The API from the comment in 2015 still makes sense to me:

``` javascript
[...]

PR #696, opened 2015-11-23
Feature: Selection Modes

musicist288:
This is a proposed implementation for https://github.com/danvk/dygraphs/issues/371.

Selection modes are defined in `src/dygraph-selection-modes.js` with the
ability to override them on a per graph basis with the option
`selectionMode` when instantiating the Dygraph.

>This is a generalized approach of the changes I needed to make for my company's use case. These changes represent the bare minimum; I wasn't sure if I should go as far as to define a whole selection model implementation (similar to the plug-in system). Our use case revolves exclusively around time series data so there may be some decisions made that are influenced by that case and may not fit the general approach.

[Review on Reviewable](https://reviewable.io/reviews/danvk/dygraphs/696)

### DIFF
--- a/auto_tests/tests/select_closest_x.js
+++ b/auto_tests/tests/select_closest_x.js
@@ -1,0 +1,110 @@
+/**
+ * @license
+ * Copyright 2015 Dan Vanderkam (danvdk@gmail.com)
+ * MIT-licensed (http://opensource.org/licenses/MIT)
+ */
+
+/**
+ * @fileoverview Test the selectByClosestX selection mode.
+ * @author musicist288@gmail.com (Joseph Rossi)
+ */
+
+import Dygraph from '../../src/dygraph';
+import * as utils from '../../src/dygraph-utils';
+
+describe("selection by closest x", function() {
+
+cleanupAfterEach();
+
+it('selects the closest previous value', function () {
+  var data = 'X,A,B\n' +
+             '100, ,2\n' +
+             '200,3, \n';
+
+  var calls = [];
+  function cb(g, seriesName, canvasContext, cx, cy, color, pointSize, idx) {
+    calls.push(arguments);
+    utils.Circles.DEFAULT.apply(this, arguments);
+  }
+
+  var g = new Dygraph(document.getElementById("graph"), data, {
+    drawHighlightPointCallback: cb,
+    selectionMode: Dygraph.SelectionModes.selectByClosestX
+  });
+
+  g.setSelection(1);
+  assert.equal(calls.length, 2);
+  var expectedA = g.toDomCoords(200, 3);
+  var aCall = calls[0];
+  assert.equal(aCall[1], 'A');
+
+  // Some implementation detail results in point.canvasy
+  // and g.toDomYCoord(point.yval) to be ever so slightly
+  // different, but not enough to be significant.
+  var acceptableError = 1e-9;
+  var difference = Math.abs(aCall[4] - expectedA[1]);
+  assert.isTrue(difference < acceptableError);
+  difference = Math.abs(aCall[3] - expectedA[0]);
+  assert.isTrue(difference < acceptableError);
+
+  var expectedB = g.toDomCoords(100, 2);
+  var bCall = calls[1];
+  assert.equal(bCall[1], 'B');
+
+  difference = Math.abs(bCall[3] - expectedB[0]);
+  assert.isTrue(difference < acceptableError);
+
+  difference = Math.abs(bCall[4] - expectedB[1]);
+  assert.isTrue(difference < acceptableError);
+});
+
+it('does not select from rows greater than the closest row', function () {
+  var data = 'X,A,B\n' +
+             '1, , \n' +
+             '2,3, \n' +
+             '3,4,5\n';
+
+  var calls = [];
+  function cb(g, seriesName, canvasContext, cx, cy, color, pointSize, idx) {
+    calls.push(arguments);
+    utils.Circles.DEFAULT.apply(this, arguments);
+  }
+
+  var g = new Dygraph(document.getElementById("graph"), data, {
+    drawHighlightPointCallback: cb,
+    selectionMode: Dygraph.SelectionModes.selectByClosestX
+  });
+
+  g.setSelection(1);
+  assert.equal(calls.length, 1);
+  assert.equal(calls[0][1], 'A');
+});
+
+it('uses the greatest xval of the selected points', function () {
+  var selectCalls = [];
+  var plugin = function () {}
+  plugin.prototype.activate = function () {
+    return {
+      select: function(e) {
+        selectCalls.push(e);
+      }
+    };
+  };
+
+  var data = 'X,A,B\n' +
+             '1, , \n' +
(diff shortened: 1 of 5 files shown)


danvk:
Thanks for this change! I'll make an effort to review it this week.

danvk:
I left a few smaller comments, but overall I'm not sure I understand what's going on. I thought the idea of this change was to support both "select by closest x-distance" and "select by closest euclidean distance," but the latter doesn't seem to be a part of this change. What's the difference between `selectByRow` and `selectByClosestX`?

---

Review status: 0 of 5 files reviewed at latest revision, 7 unresolved discussions.

---

*[src/dygraph-options-reference.js, line 854 \[r1\]](https://reviewable.io:443/reviews/danvk/dygraphs/696#-K4Ol6xBvSN1YjEHiNNE:-K4Ol6xBvSN1YjEHiNNF:1072190256) ([raw file](https://github.com/danvk/dygraphs/blob/cc5be8851bf1c49bff79d17ae32a443dc438389d/src/dygraph-options-reference.js#L854)):*
closest (sp)

---

*[src/dygraph-options-reference.js, line 855 \[r1\]](https://reviewable.io:443/reviews/danvk/dygraphs/696#-K4OkvarRdE5UUpRCEux:-K4OkvarRdE5UUpRCEuy:1485669668) ([raw file](https://github.com/danvk/dygraphs/blob/cc5be8851bf1c49bff79d17ae32a443dc438389d/src/dygraph-options-reference.js#L855)):*
I'd like to move away from passing the dygraph as a positional parameter to callbacks. It should be accessed as `this`.

---

*[src/dygraph-options-reference.js, line 858 \[r1\]](https://reviewable.io:443/reviews/danvk/dygraphs/696#-K4Oku-CwXxIg65YsftO:-K4Oku-CwXxIg65YsftP:178526072) ([raw file](https://github.com/danvk/dygraphs/blob/cc5be8851bf1c49bff79d17ae32a443dc438389d/src/dygraph-options-reference.js#L858)):*
`It's` → `Its`

---

*[src/dygraph-selection-modes.js, line 8 \[r1\]](https://reviewable.io:443/reviews/danvk/dygraphs/696#-K4OksfZ1b-nM3l7mcNf:-K4OksfZ1b-nM3l7mcNg:-1455997968) ([raw file](https://github.com/danvk/dygraphs/blob/cc5be8851bf1c49bff79d17ae32a443dc438389d/src/dygraph-selection-modes.js#L8)):*
`fileoverview`

---

*[src/dygraph-selection-modes.js, line 10 \[r1\]](https://reviewable.io:443/reviews/danvk/dygraphs/696#-K4OlD6-pyEpSDGoGNWt:-K4OlD6-pyEpSDGoGNWu:-1531186233) ([raw file](https://github.com/danvk/dygraphs/blob/cc5be8851bf1c49bff79d17ae32a443dc438389d/src/dygraph-selection-modes.js#L10)):*
`dygraph.js`

---

*[src/dygraph-selection-modes.js, line 17 \[r1\]](https://reviewable.io:443/reviews/danvk/dygraphs/696#-K4OlH9jAoKDeM8P8TOo:-K4OlH9jAoKDeM8P8TOp:-1033239155) ([raw file](https://github.com/danvk/dygraphs/blob/cc5be8851bf1c49bff79d17ae32a443dc438389d/src/dygraph-selection-modes.js#L17)):*
"that strictly" → did a word get dropped here?

---

*[src/dygraph.js, line 3578 \[r1\]](https://reviewable.io:443/reviews/danvk/dygraphs/696#-K4Om0TeA8DGi3vAD1wk:-K4Om0TeA8DGi3vAD1wl:-2102066962) ([raw file](https://github.com/danvk/dygraphs/blob/cc5be8851bf1c49bff79d17ae32a443dc438389d/src/dygraph.js#L3578)):*
any reason not to just do:

```javascript
Dygraph.SelectionModes = DygraphSelectionModes;
```

?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/danvk/dygraphs/696#-:-K4OmFt-krxzN4I4dAts:665830134)*

musicist288:
The difference is that with selectByRow, if the selected row does not have a value for a series, the selection will not include a value for that series. selectByClosestX starts at the current row, but if no value is found for a series, it looks for the last defined value for that series in row numbers less than the selected row. Perhaps a more appropriate name would be something like "selectByLatestDefinedValue"?

Perhaps I let my needs dilute my understanding of the original feature request  I'm exclusively visualizing time-series data. One chart type in particular has data where a single row would almost never have defined values for every series. My goal was to get the highlight dots and the legend reporting the last value reported for the series up until the current time, but never after, and have the legend report the time value of the latest data point.

Here's a [GIF](https://drive.google.com/file/d/0B0ijexLSVGAFcThNamw1MUo4MDg/view?usp=sharing) of the interaction I needed.  This is the most extreme case, but I have other examples if needed.

I'm not too concerned about this specific selection model making it into the library if we can't agree about the semantics of the selection mode. I'm mostly concerned about the callback option being exposed.

---

Review status: 0 of 5 files reviewed at latest revision, 3 unresolved discussions.

---

*[src/dygraph-options-reference.js, line 855 \[r1\]](https://reviewable.io:443/reviews/danvk/dygraphs/696#-K4OkvarRdE5UUpRCEux:-K4P5nnvqysyf3PFrdGy:918055650) ([raw file](https://github.com/danvk/dygraphs/blob/cc5be8851bf1c49bff79d17ae32a443dc438389d/src/dygraph-options-reference.js#L855)):*
Ok. I wasn't sure about that. I'll change it.

---

*[src/dygraph-selection-modes.js, line 17 \[r1\]](https://reviewable.io:443/reviews/danvk/dygraphs/696#-K4OlH9jAoKDeM8P8TOo:-K4OzrqV7QZa57BsQ7p4:1259265844) ([raw file](https://github.com/danvk/dygraphs/blob/cc5be8851bf1c49bff79d17ae32a443dc438389d/src/dygraph-selection-modes.js#L17)):*
Not dropped, but I was editing that block a lot. I'll re-word it.

---

*[src/dygraph.js, line 3578 \[r1\]](https://reviewable.io:443/reviews/danvk/dygraphs/696#-K4Om0TeA8DGi3vAD1wk:-K4P-77hgGGLzNdAP-fg:-2118186699) ([raw file](https://github.com/danvk/dygraphs/blob/cc5be8851bf1c49bff79d17ae32a443dc438389d/src/dygraph.js#L3578)):*
Nope, not sure why I originally had this. I'll change it to an assignment.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/danvk/dygraphs/696#-:-K4P0D81eyBbQ_lPtZdm:1694094202)*

danvk:
I'd accept a version of this if I saw a demonstration that it could be used to fix #371, i.e. support selecting the closest point by euclidean distance. Ideally that would share code with the existing mechanism via `highlightSeriesOpts`.

musicist288:
Ok. I'll see if I can do that. I'm, probably just going to maintain a
separate fork for now

musicist288:
Closing for now. Will reopen if a fix for #371 is implemented.

PR #703, opened 2015-12-08
[Feature Request] Provide option to set color and width for annotation line

crazyrohila:
I haven't find a way to set color and width for annotation line, I have added option to set that.

PS. I was not sure against which branch I should create PR, so created against master.

### DIFF
--- a/docs/annotations.html
+++ b/docs/annotations.html
@@ -192,6 +192,8 @@ g.ready(function() {
 <tr><td><code>height</code></td><td>Height (in pixels) of the annotation flag or icon.</td></tr>
 <tr><td><code>cssClass</code></td><td>CSS class to use for styling the annotation.</td></tr>
 <tr><td><code>tickHeight</code></td><td>Height of the tick mark (in pixels) connecting the point to its flag or icon.</td></tr>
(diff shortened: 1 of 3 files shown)


danvk:
These should be documented [here][1].

Have you tried the [super annotations plugin][2]? I believe it uses a `<div>` for the connector, which would let you set these sorts of styles with CSS. http://dygraphs.com/tests/hairlines.html

[1]: https://github.com/danvk/dygraphs/blob/dd74404fac69f23183ff8e540a018ca35bab4f7a/docs/annotations.html#L185
[2]: https://github.com/danvk/dygraphs/blob/dd74404fac69f23183ff8e540a018ca35bab4f7a/src/extras/super-annotations.js

crazyrohila:
Thanks Dan.

Super annotations plugin also provide whole bunch of other features, which seems like overkill if we want to just change the visual of annotation line. It seems like useful addition if we can provide these options with base annotation plugin.

Added documentation for the new options.

danvk:
I'm going to merge this. These aren't top-level options, so I feel less biased against adding new ones.

PR #678, opened 2015-10-25
Move styles over to CSS

danvk:
This creates `dygraph.css` and moves whatever styles we can into it.

The styles which are hard to move set position properties like `left` and `top`. I may wind up creating a `div` which sits precisely on top of the chart area. This would allow `left: 0` to and `right: 0` to be meaningful in the presence of chart labels.

Remaining tasks:

- [x] Add CSS file to all demos and the gallery.
- [x] Kill `labelsDivWidth`
- [x] Document how to replace each option that was removed.
- [x] grep for removed options
- [x] styled-chart-labels gallery entry is broken.
- [ ] `:undefined` on highlight-closest-series gallery (this is also broken at HEAD)

Options which are removed:

- `labelsDivWidth`
    Replace with `.dygraph-legend { width: 100px; }`

- `labelsDivStyles`
    Replace with `.dygraph-legend { ... }`

- `axisLabelColor`
    Replace with `.dygraph-axis-label { color: ... }` or `.dygraph-axis-label-x { color: ... }`.

Fixes #508

PR #711, opened 2016-01-05
Custom Zoom Handle Image

blenderous:
Hi there,

This pull request is for the new option that I have added to dygraphs.

It's called `rangeSelectorZoomHandleImage`.

It is to add a custom zoom handle image for the Range Selector.

screenshot from 2016-01-05 11 46 55 custom zoom handle image

I am not able to figure out why the code does not pass the `range-selector-interaction` test through.

If you could please help me figure it out, it would be awesome.

:metal:

Thank you.


danvk:
This should be done via CSS. See #678.

PR #717, opened 2016-01-25
Bugfix for the zoom rectangle and legend on dynamic updates

1r0b1n0:
On the dynamic-update example, the grey zoom rectangle disappears on each update.
The legend has the same problem : on an update the legend disappears (but reappears on a mouse move event).
This pull request fixes both issues.


jaredscheib:
Hi there,

We use `dygraphs` extensively and have this same issue of legends disappearing on dynamic update (`.updateOptions`).

I've noticed there are many PRs and issues outstanding -- do you know if/when this might be merged?

PR #718, opened 2016-01-25
Additional granularities and tick placements (2yr, 5yr)

fungl164:


fungl164:
Two separate features.  Should I resubmit?

fungl164:
Submitting as separate

PR #719, opened 2016-01-25
Div-wrapping axis labels for better DOM sectioning

fungl164:
This provides better DOM separation for dygraph axes and canvases.

### DIFF
--- a/src/plugins/axes.js
+++ b/src/plugins/axes.js
@@ -27,6 +27,9 @@ Options left to make axis-friendly.
 var axes = function() {
   this.xlabels_ = [];
   this.ylabels_ = [];
+  this.containerDivX = document.createElement('div');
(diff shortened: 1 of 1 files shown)


danvk:
You're going to need to do a better job of explaining why this is helpful.

fungl164:
Not critical, but while investigating why there is a bleed of selected points when mousing over the y-axis labels, I started to look at the DOM arrangement and saw that all the labels are appended as direct children of the same node that the canvases are in.  

Even if absolutely positioned, this separates the x, y, y2 labels in the DOM into their own divs to make it more clear to me where things belong given labels are managed through the DOM vs painted straight onto canvases...

Hope that makes sense... : )

danvk:
I don't really see a lot of upside to this, to be honest.

PR #721, opened 2016-01-25
Pass seriesData canvasx + canvasy point coordinates to legendFormatter

fungl164:
This allows legendFormatter to make use of CSS positioning for individual series data.

### DIFF
--- a/src/plugins/legend.js
+++ b/src/plugins/legend.js
@@ -268,6 +268,10 @@ Legend.generateLegendHTML = function(g, x, sel_points, oneEmWidth, row) {
       var seriesData = labelToSeries[pt.name];
       seriesData.y = pt.yval;
 
+      // Add canvas location to returned series data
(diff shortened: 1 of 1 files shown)


danvk:
Can you give an example of why you'd want this information? It's not clear to me why you'd need pixel information when formatting text for the legend.

fungl164:
This will allow something similar to the legend 'follow' option, but for individual series points.  Using absolute positioning, I can place legend items to track along the points of each plotted line or (if I wanted to) to track only the y-movements by fixing them to a the side (right or left).

Makes sense?

danvk:
Sorry this stalled. I think I'd be fine merging this PR if it included an example of what you describe ("something similar to the legend 'follow' option, but for individual series points").

danvk:
I'm going to close this out. Feel free to reopen with an example.

PR #725, opened 2016-02-08
Bugfix: Import missing dygraph-utils.js

### LINKED ISSUE #724: Uncaught ReferenceError: utils is not defined when page contains iframes
In 2cfded32d5da38aad8d9617654569f3e3f1c420d the IFrameTarp functionality was split out to a separate file. 
However that file does not import the `dygraphs-utils.js` and thus the call `var pos = utils.findPos(iframe)`  will throw an `Uncaught ReferenceError` on [line 39](https://github.com/danvk/dygraphs/blob/master/src/iframe-tarp.js#L39)

timeu:
The IFrameTarp.prototype.cover function calls `findPos` from the `dygraph-utils.js` file but does not import it. 
This fixes #724

### DIFF
--- a/src/iframe-tarp.js
+++ b/src/iframe-tarp.js
@@ -23,6 +23,8 @@
  *
  * @constructor
  */
+import * as utils from './dygraph-utils';
+
 function IFrameTarp() {
   /** @type {Array.<!HTMLDivElement>} */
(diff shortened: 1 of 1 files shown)


danvk:
Thanks!

Issue #745, opened 2016-05-05
With "drawGapEdgePoints", unwanted point often drawn at beginning of chart


mrcslws:
Here's a repro: http://jsfiddle.net/bt3fdkkg/

EXPECTED:
No points are drawn

ACTUAL:
A point is drawn

~~~javascript

new Dygraph(
  document.getElementById("demodiv"),
  [[0, 0],
   [1, 1],
   [2, 2],
   [3, 3],
   [4, 4],
   [5, 5]],
  {drawGapEdgePoints: true,
   pointSize: 5,
   axes: { y: { drawAxis: false } }});

~~~

It happens when the first point is at `canvasx` of 0 and `drawGapEdgePoints` is enabled.

I have a fix.

PR #746, opened 2016-05-05
Bugfix: Unwanted drawn point when prevCanvasX is 0

mrcslws:
Fixes issue #745

---
This change is [Reviewable](https://reviewable.io/reviews/danvk/dygraphs/746)


danvk:
I'll merge this if you add a unit test (the code you included in the issue is a good start).

mrcslws:
Sounds good. Is this one good? I verified that the second one fails without my commit, and succeeds with it.

danvk:
Beautiful, thanks for the fix and the test!

PR #538, opened 2015-01-28
Candlestick plotter

przmv:
Candlestick chart with optional bars compression annually, quarterly, monthly, weekly or daily.

PR #758, opened 2016-06-30
Add option for drawing box plots

vhotspur:
The PR adds `boxplot: true` option to plot the data as [box plots](https://en.wikipedia.org/wiki/Box_plot). The data has to be precomputed, i.e. user specifies values for the whiskers and the box, not raw samples (outliers are not supported).

The data are expected to have 5 (median, first and third quartile and position of the whiskers) or 6 values (if the point is not median but average, for example).

An example was added to the gallery to demonstrate this feature.


danvk:
Is this different than the candlestick plotter in http://dygraphs.com/tests/plotters.html and #538?

vhotspur:
> Is this different than the candlestick plotter in http://dygraphs.com/tests/plotters.html and #538? 

Although they are visually similar, there are several differences. Boxplot allows you to display basic properties of the distribution behind any generic data, candlestick chart targets financial community only (fixed meaning of colors and shadows). A boxplot requires 5 values to be drawn (the patch do not handle outliers), candlestick needs only 4 (it misses the middle line with median). And there are also some subtle visual differences (the median line, shadow vs whisker).

The values displayed by the boxplot can be probably somehow emulated by the candlestick chart and extra serie for the medians. But there would be still the small visual difference that complicates "reading" the graph.

vhotspur:
@danvk what do you think about this PR? Do I need to document it more, add more tests or something else is missing? Thanks!

PR #761, opened 2016-07-08
Check for new options before updating synchronised graphs (Fixes #760)

### LINKED ISSUE #760: Synchronizer causes excessive redraws
Upon resizing a window that containing a bunch of synchronized graphs, I noticed the performance of Dygraphs wasn't too great. Looking a bit further into this I found that each graph is redrawn multiple times for a single resize event (e.g. maximise/minimise the browser window).

It seems that the `drawCallback` on each graph setup by the synchronizer calls `updateOptions` on every other graph regardless of whether there are actually new options or not, and this causes a storm of calls to `renderGraph_` and slows things down.

See: http://jsfiddle.net/qcu2xd3z/

Version: 1.1.1 & dev

justinsg:
Fixes issue #760 by checking for new options before calling `updateOptions` with an extra flag.

### DIFF
--- a/src/extras/synchronizer.js
+++ b/src/extras/synchronizer.js
@@ -153,6 +153,16 @@ var synchronize = function(/* dygraphs..., opts */) {
   };
 };
 
+function arraysAreEqual(a, b) {
+  if (!Array.isArray(a) || !Array.isArray(b)) return false;
+  var i = a.length;
+  if (i !== b.length) return false;
+  while (i--) {
+    if (a[i] !== b[i]) return false;
+  }
+  return true;
+}
+
 function attachZoomHandlers(gs, syncOpts, prevCallbacks) {
   var block = false;
   for (var i = 0; i < gs.length; i++) {
@@ -173,6 +183,13 @@ function attachZoomHandlers(gs, syncOpts, prevCallbacks) {
             }
             continue;
           }
+
+          // Only redraw if there are new options
+          if (arraysAreEqual(opts.dateWindow, gs[j].getOption('dateWindow')) && 
+              arraysAreEqual(opts.valueRange, gs[j].getOption('valueRange'))) {
+            continue;
+          }
+
           gs[j].updateOptions(opts);
         }
         block = false;


src/extras/synchronizer.js
```diff
@@ -173,7 +183,13 @@ function attachZoomHandlers(gs, syncOpts, prevCallbacks) {
...
             continue;
           }
-          gs[j].updateOptions(opts);
+
+          // Only redraw if there are new options
+          var blockRedraw = true;
+          if (!arraysAreEqual(opts.dateWindow, gs[j].getOption('dateWindow'))) blockRedraw = false;
+          else if (!arraysAreEqual(opts.valueRange, gs[j].getOption('valueRange'))) blockRedraw = false;
+
+          gs[j].updateOptions(opts, blockRedraw);
```
danvk: Why use `blockRedraw` rather than not calling `updateOptions` at all?

justinsg:
> Why use `blockRedraw `rather than not calling `updateOptions `at all?

Good point @danvk - have updated the PR to skip `updateOptions` instead

src/extras/synchronizer.js
```diff
@@ -153,6 +153,16 @@ var synchronize = function(/* dygraphs..., opts */) {
   };
 };
 
+function arraysAreEqual(a, b) {
+  if (!Array.isArray(a) || !Array.isArray(b)) return false;
+  var i = a.length;
+  if (i != b.length) return false;
```
danvk: Use `!==` here, too.

src/extras/synchronizer.js
```diff
@@ -173,6 +183,14 @@ function attachZoomHandlers(gs, syncOpts, prevCallbacks) {
             }
             continue;
           }
+
+          // Only redraw if there are new options
+          if (
+            arraysAreEqual(opts.dateWindow, gs[j].getOption('dateWindow')) &&
```
danvk: Style nit: move this line up on the same line as `if`.

src/extras/synchronizer.js
```diff
@@ -173,6 +183,14 @@ function attachZoomHandlers(gs, syncOpts, prevCallbacks) {
             }
             continue;
           }
+
+          // Only redraw if there are new options
+          if (
+            arraysAreEqual(opts.dateWindow, gs[j].getOption('dateWindow')) && 
+            arraysAreEqual(opts.valueRange, gs[j].getOption('valueRange'))
+          )
```
danvk: Style nit: add curlies for multiple line if statements.

(I try to follow https://google.github.io/styleguide/javascriptguide.xml)

danvk: move the `)` up on the same line as the second `arraysAreEqual`.

danvk:
A couple very small style nits. Otherwise we're good to go. Please squash your commits & I'll merge.

danvk:
Perfect, thanks!

PR #767, opened 2016-08-11
Added rangeSelectorBounds option

neurot1k:
Limits the range selector so that a section of the data is always visible.

### DIFF
--- a/auto_tests/tests/range_selector.js
+++ b/auto_tests/tests/range_selector.js
@@ -21,11 +21,9 @@ cleanupAfterEach();
 var restoreConsole;
 var logs = {};
 beforeEach(function() {
-  restoreConsole = Util.captureConsole(logs);
 });
 
 afterEach(function() {
-  restoreConsole();
 });
 
 it('testRangeSelector', function() {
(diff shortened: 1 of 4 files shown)


danvk:
Does this control the x-axis range of the range selector? Why would you want to show less than the full domain of the data?

auto_tests/tests/range_selector.js
```diff
@@ -21,11 +21,11 @@ cleanupAfterEach();
 var restoreConsole;
 var logs = {};
 beforeEach(function() {
-  restoreConsole = Util.captureConsole(logs);
+  //restoreConsole = Util.captureConsole(logs);
```
LM450N: Could you remove the useless comments please?

neurot1k:
Yes, it limits the x-axis range. Why would you want this? Perhaps you have some highlighted data that you always want displayed. 

Specifically, I'm using the range selector to create a segment of a larger gps track and annotations to define waypoints. Once a waypoint is defined it must be a part of the segment so I want to limit the range that can be selected. 

I understand this may be a specialized case, so if you don't think it adds any value to the project, no worries.

auto_tests/tests/range_selector.js
neurot1k: Thanks, those have been removed.

danvk:
Yeah, I don't see this as a general use case. If you want to make only a portion of your data accessible in the chart, you're better off filtering the data before you give it to dygraphs.

Thanks for your contribution, though!

LM450N:
_[my two cents]_ 💸 
This feature could prevents to reach a white screen when the user move the graph where there is no data?


move-graph-out-range

PR #770, opened 2016-08-25
Fix independent series doc

Zacaria:
In [this](http://dygraphs.com/gallery/#g/independent-series) doc page, the second chart did not display data.

JavaScript was throwing `Couldn't parse 1 as a date` for each csv line data.

For the sake of simplicity I switched to array notation and now the graph displays well.

### DIFF
--- a/gallery/independent-series.js
+++ b/gallery/independent-series.js
@@ -145,16 +145,18 @@ Gallery.register(
 
     new Dygraph(
       document.getElementById('graph2'),
-      'x,A,B  \n' +
-      '1,,3   \n' +
-      '2,2,   \n' +
-      '3,,5   \n' +
-      '4,4,   \n' +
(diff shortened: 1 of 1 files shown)


danvk:
In which browser? I see the second chart on http://dygraphs.com/gallery/#g/independent-series in Chrome.

Zacaria:
In firefox 48,
The bug happens on the site and on jsfiddle.

I confirm that it works on Chrome.

danvk:
Can confirm this is broken in FF. Could you try to track down why, and whether this broke recently (i.e. in a recent FF)? While your change does fix the issue, it worries me that `1` is being parsed as a date.

danvk:
I'm going to close this out due to inactivity. Feel free to reopen if you'd like to track it down a bit more.

Issue #637, opened 2015-06-29
Series names beyond last visible series are inaccessible


ddraayer:
If I have series A, B, and C, and only A is visible, I would like to be able to make B visible via a g.setVisibility call. To get the index for series B requires a g.indexFromSetName("B") call, but this yields undefined. It appears that indexFromSetName returns undefined for any series name beyond the last visible series. See https://jsfiddle.net/eM2Mg/6479/ for an example.

I took only a quick peek at the code, but I suspect that in the definition of Dygraph.prototype.drawGraph_ in dygraph.js, a reference to `points.length` should be `labels.length` instead (?):
```javascript
for (var i = 1; i < points.length; i++) {
    this.setIndexByName_[labels[i]] = i;
    if (!this.visibility()[i - 1]) continue;
    this.layout_.addDataset(labels[i], points[i]);
    this.datasetIndex_[i] = dataIdx++;
}
```

danvk:
That is odd. My understanding was that `g.getValue(0, indexFromSetName(x)) == x` should be true for every column.

The solution would be to fix `indexFromSetName`. Not generating intermediate data for invisible series is a nice optimization that I wouldn't want to drop.

danvk:
Fixed by #771

Issue #434, opened 2014-10-20
setVisibility() can cause the last data series to vanish from setIndexByName_


danvk:
_From [sergeysl...@gmail.com](https://code.google.com/u/108873442384158821856/) on July 09, 2014 17:45:30_

Here's a JSFiddle reproducing the problem: http://jsfiddle.net/X68vR/2/ Here's how to reproduce this error in the JSFiddle:

1) Use the checkbox to hide the last time series (Random2).

2) Wait one second for the graph to update.

3) Attempt to show the last time series using the checkbox.

Expected:

Time series should be shown again.

Actual:

Time series is not shown again, because indexFromSetName() can't find the series anymore, because it's gone from setIndexByName_

This bug does not happen on any time series except the very last one.

This bug also doesn't happen if you don't use indexFromSetName(), and just reference the series id directly.

_Original issue: http://code.google.com/p/dygraphs/issues/detail?id=518_

sslepian:
As a note, getPropertiesForSeries() can also be used to get the series index (it's the column property) and does not suffer from this issue.

danvk:
Fixed by #771

PR #771, opened 2016-08-25
Bug fix: indexFromSetName on invisible sets

squaregoldfish:
Calling `indexFromSetName` on the last set would fail if its visibility was set to false, because the set's label would not be added to the internal list of sets (stored in the `setIndexByName_` variable).

This pull request fixes the issue, and adds a test to check that it works.

The test added to `tests/data_api.js` provides an example that currently fails, and now passes with this fix.

### DIFF
--- a/auto_tests/tests/data_api.js
+++ b/auto_tests/tests/data_api.js
@@ -4,6 +4,7 @@
  * @author danvdk@gmail.com (Dan Vanderkam)
  */
 import Dygraph from '../../src/dygraph';
+import * as utils from '../../src/dygraph-utils';
 describe("data-api", function() {
 
 cleanupAfterEach();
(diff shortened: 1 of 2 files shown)


auto_tests/tests/data_api.js
```diff
@@ -99,4 +99,21 @@ it('testGetRowForXDuplicates', function() {
   assert.equal(5, g.getRowForX(9));
 });
 
+// indexFromSeriesName should return a value even if the series is invisible
+// In 1.1.1, if you request the last set and it's invisible, the method returns undefined.
+it('testIndexFromSetNameOnInvisibleSet', function() {
+  
+  var localOpts = opts;
```
danvk: This doesn't actually copy anything. If you want to do that there's a method in utils.

danvk:
Other than my one comment, this looks great. Thanks for the fix!

Will this address #434 and #637?

squaregoldfish:
Yes, I think it addresses both those issues.

I'll take a look at the test and updated it properly tomorrow.

danvk:
Squashed & merged. Thanks!

squaregoldfish:
NP. Don't forget about #434 and #637 :)

danvk:
Thanks for the reminder! Closed them out.

PR #773, opened 2016-09-09
Check that setRow is not negative, fixes #644.

### LINKED ISSUE #644: Mouseover on synchronized graph without zoom causes error
Can be seen in http://dygraphs.com/gallery/#g/synchronize that when you uncheck **Zoom** (with selection option active), and zoom on one of the graphs, the other graphs aren't synced (as expected). But after hovering over one of the other (unzoomed) graphs, it will randomly throw back an undefined error within `Dygraph.prototype.setSelection`.

Majlen:
Synchronization breaks up if this case is not checked. Fix has been
proposed in the bugreport.


danvk:
Thanks!

Issue #776, opened 2016-09-16
Allow millisecond date ticks


jtbandes:
Hi, I just started using dygraphs and it's great! 😄 

I noticed that dygraph-tickers.js seems to [define `DATEFIELD_MS`](https://github.com/danvk/dygraphs/blob/b4b9eeac8b116c18bbf6ebb0c4d542a3f8d2ce32/src/dygraph-tickers.js#L257), but this value is not used in any of the `TICK_PLACEMENT`s. It'd be great if Granularity had some values like `{FIVE,TWO,ONE}_HUNDRED_MILLISECONDLY`, `{FIFTY,TWENTY,TEN}_MILLISECONDLY`, and `{FIVE_,TWO_,}MILLISECONDLY` whose placements could use the milliseconds field.

On a broader note, though, as a client of the library, I wasn't able to do this myself very easily because of how the `dateTicker` is structured — even though `getDateAxis` is exported, the `granularity` passed to it is used to look up a value in the non-exported `TICK_PLACEMENT` array (and `pickDateTickGranularity` is also private). This might be improved by having `getDateAxis` accept the `{datefield, step, spacing}` as arguments, rather than the granularity — then users could replicate+modify the default `dateTicker` more easily. Just a thought 😺

jtbandes:
(I'd be up for implementing this and submitting a PR, but you'll have to let me know which approach you prefer.)

pierriko:
Hi, I had started working on that but din't finish yet, I just opened https://github.com/danvk/dygraphs/pull/777

jtbandes:
Looks like this was done in https://github.com/danvk/dygraphs/pull/893

PR #777, opened 2016-09-22
[tickers] add milliseconds granularity

PierrickKoch:
Still needs some tests in `auto_tests/tests/date_ticker.js`, and maybe cleaner display in `src/dygraph-utils.js` (don't show milliseconds if granularity > MINUTELY or so).


src/dygraph-tickers.js
```diff
@@ -220,29 +220,36 @@ export var dateTicker = function(a, b, pixels, opts, dygraph, vals) {
...
-  DECADAL: 20,
-  CENTENNIAL: 21,
-  NUM_GRANULARITIES: 22
+  MSLY: 0,
+  TWO_MSLY: 1,
+  FIVE_MSLY: 2,
+  TEN_MSLY: 3,
+  FIFTY_MSLY: 4,
+  HTH_MSLY: 5,
+  FHTH_MSLY: 6,
```
jtbandes: What does HTH mean? At first I though "hundred thousand", but that doesn't make sense here; this is just 100/500.

pierriko: My bad, you're right :)
Would `HUNDRED_MSLY` and `FIVE_HUNDRED_MSLY` be ok for you ?

jtbandes: Personally, I think there's no harm in spelling out "millisecond" instead of just using "ms". But, it's not my code 😉  (Also, as I mentioned in #776 I think it'd be interesting to explore alternative approaches that make this more customizable.)

auto_tests/tests/date_ticker.js
```diff
@@ -263,6 +263,7 @@ it('testAllDateTickers', function() {
   assert.deepEqual([{"v":1307908020000,"label":"19:47"},{"v":1307908050000,"label":"19:47:30"},{"v":1307908080000,"label":"19:48"},{"v":1307908110000,"label":"19:48:30"},{"v":1307908140000,"label":"19:49"},{"v":1307908170000,"label":"19:49:30"}], ticker(1307908000112, 1307908172259, 800, options));
   assert.deepEqual([{"v":1307908020000,"label":"19:47"},{"v":1307908050000,"label":"19:47:30"},{"v":1307908080000,"label":"19:48"},{"v":1307908110000,"label":"19:48:30"},{"v":1307908140000,"label":"19:49"},{"v":1307908170000,"label":"19:49:30"}], ticker(1307908000112, 1307908173260, 800, options));
   assert.deepEqual([{"v":978307200000,"label":"Jan 2001"},{"v":986083200000,"label":"Apr 2001"},{"v":993945600000,"label":"Jul 2001"},{"v":1001894400000,"label":"Oct 2001"}], ticker(978307200000, 1001894400000, 400, options));
+  // TODO add millisecond test ?
```
danvk: yes please.

src/dygraph-tickers.js
```diff
@@ -220,29 +220,36 @@ export var dateTicker = function(a, b, pixels, opts, dygraph, vals) {
...
-  TWO_DAILY: 14,
-  WEEKLY: 15,
-  MONTHLY: 16,
-  QUARTERLY: 17,
-  BIANNUAL: 18,
-  ANNUAL: 19,
-  DECADAL: 20,
-  CENTENNIAL: 21,
-  NUM_GRANULARITIES: 22
+  MILLISECONDLY: 0,
```
danvk: I'd prefer to avoid renumbering existing granularities. You can put `MILLISECONDLY` first in the list, but make its number 22.

src/dygraph-utils.js
```diff
@@ -1229,6 +1229,11 @@ export function dateAxisLabelFormatter(date, granularity, opts) {
     if (frac === 0 || granularity >= DygraphTickers.Granularity.DAILY) {
       // e.g. '21 Jan' (%d%b)
       return zeropad(day) + '&#160;' + SHORT_MONTH_NAMES_[month];
+    } else if (granularity < DygraphTickers.Granularity.SECONDLY) {
+      var str = "" + millis;
+      return zeropad(secs) + "." + ('000'+str).substring(str.length);
```
danvk: Add an example of what this looks like.

danvk (changes requested):
Thanks for making this change! I'd like to accept but will need to see a unit test (in `auto_tests`).

auto_tests/tests/date_ticker.js
121watts: @danvk if I write a test for this, would that get this over the line for a merge?

src/dygraph-tickers.js
mirabilos: Is that wise? Not renumbering breaks less/greater-than relationships in total ordering… I’d go for it, especially as it’s this once only (nothing finer than ms supported in js)…

PR #787, opened 2016-10-01
Remove some dead data handler coder

danvk:


danvk:
cc @davideberlein unless I'm missing something here? These functions are never called and are marked `@private`.

davideberlein:
Seems fine to me to remove those

Issue #751, opened 2016-06-02
[ "clear" problem] When changing YAxis from 1 to 2 by using updateOptions.


flexdeviser:
first I created a single yAxis graph, after that I update the options to two yAxis, then I got this graph.
![image](https://cloud.githubusercontent.com/assets/3424332/15733254/794c69ee-28c7-11e6-929e-3830a0138a2f.png)
The right side of graph has something wrong. so I think, you need to clear the canvas before drawing 'y2' axis.
btw, I just add "this.plotter_.clear();" into "Dygraph.prototype.updateOptions", and the problem fixed.

jpardey:
Thanks for posting this. I've been having the same problem, and your fix solved it for me.

For my own usage, I've added the `this.plotter_.clear()` call into the `if (!block_redraw)` block within the `if (file)` block of that function. Is there a better place to include it? I doubt this will be an issue for my work, but I could imagine it still being a mess if I was changing axis settings without updating the files options, and I have no idea how it will work with annotations.

flexdeviser:
@jpardey I'm so glad your problem solved. But I think "this.plotter_.clear()" is just a temporary solution. I expect there will be a good way or some guy to solve this problem.

danvk:
Saem as #629

Issue #629, opened 2015-06-15
y2 axis does not render correctly after updateOptions


FrEaKmAn:
I have an option where users can define series axis (y or y2). When there is no value defined for y2 axis and I add it, it breaks the chart.

I have provided the demo: http://jsfiddle.net/9un6yxe3/

Just hit the update button. It will update series and move 'B' from y to y2 axis, but it won't render chart correctly.

danvk:
Looks like the chart area isn't getting properly cleared before being redrawn:

![screen shot 2015-06-16 at 1 22 04 pm](https://cloud.githubusercontent.com/assets/98301/8189499/b6286ab4-142a-11e5-9143-5864ddb4230f.png)

FrEaKmAn:
Any ideas how to resolve it?

rochejul:
Same issue !

It seems a bad synchronization between the axes options and the axis plugin

FrEaKmAn:
any updates about this?

WeAreElite:
I got the same issue but it had to work so i fiddled around a bit and found this: http://jsfiddle.net/9un6yxe3/
It is a crazy workaround, not at all good but if it really has to work in a short notice maybe you can give it a shot.

key10:
I'm late to the party here, but just in case others run across this on dygraphs v1.1.1 like myself, I've gotten closer the behavior expected. 
I've created a fiddle here: http://jsfiddle.net/fokmohoh/
which is a fork of the fiddle @FrEaKmAn and @WeAreElite created/worked on.

The `update` button cycles between 1 and 2 y-axes, calling `updateOptions`. The main part I changed was the double call to `updateOptions` with the `showRangeSelector` being toggled to 'kick' Dygraph, as well as explicitly declaring the `drawAxis` property for both y-axes in the constructor. 

This seems to add/remove the second y-axis without the tearing/overlap as seen in the fiddle that @FrEaKmAn created.

[...]

PR #793, opened 2016-10-06
Clear graph when numAxes goes from 1 -> 2 in updateOptions

neurot1k:
Address issue #629 and #751

### DIFF
--- a/src/dygraph.js
+++ b/src/dygraph.js
@@ -3138,6 +3138,7 @@ Dygraph.prototype.updateOptions = function(input_attrs, block_redraw) {
   // copyUserAttrs_ drops the "file" parameter as a convenience to us.
   var file = input_attrs.file;
   var attrs = Dygraph.copyUserAttrs_(input_attrs);
+  var prevNumAxes = this.attributes_.numAxes();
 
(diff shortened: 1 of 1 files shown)


src/dygraph.js
```diff
@@ -3167,6 +3168,7 @@ Dygraph.prototype.updateOptions = function(input_attrs, block_redraw) {
 
   this.attributes_.reparseSeries();
 
+  if (prevNumAxes < this.attributes_.numAxes()) this.plotter_.clear();
```
danvk: Would adding this condition to the `requiresNewPoints` check also fix the issue?

i.e.

```js
var requiresNewPoints = utils.isPixelChangingOptionList(this.attr_("labels"), attrs) ||
    prevNumAxes != this.attributes_.numAxes();
```

And does the check need to be `prevNumAxes != this.attributes_.numAxes()`?

neurot1k: Adding the condition to requiresNewPoints does not solve the problem.

prevNumAxes != this.attributes_.numAxes() will work, however if you are changing from 2 axes to 1 the graph will overwrite the second axis so calling clear() is not necessary.

danvk:
Merging since this does fix the immediate issue.

PR #805, opened 2016-12-06
add the ability to highlight multiple series with setSelection

kchengsf:
I'd like to add the ability to highlight multiple series in setSelection.

I've added a test and updated relevant documentation parts to accommodate this, however there are quite a few ways this could be implemented.  Please let me know if a different approach makes more sense.  In particular, repeatedly calling _renderLineChart is cleaner but requires iteration through all series each call.  Moving the logic downward requires the method to understand the nuance between an array vs string option.

### DIFF
--- a/auto_tests/tests/selection.js
+++ b/auto_tests/tests/selection.js
@@ -50,6 +50,20 @@ it('testSetGetSelectionDense', function() {
   assert.equal(3, g.getSelection());
 });
 
+it('testSetGetSelectionMulti', function() {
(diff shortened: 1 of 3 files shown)


danvk:
@klausw any thoughts on this? Seems fine to me other than style guide issues.

klausw:
I think this looks good, and it's functionality I've been wanting myself. I don't have a strong preference about the implementation. There may be useful followups such as restricting interactive highlighting to switch among the selected series, assuming that this isn't already working.

PR #839, opened 2017-02-08
Move typings into Dygraph library

jacobbaskin:
Let me know if you want a test or something.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,1264 @@
+// Type definitions for dygraphs 1.1.1
+declare namespace dygraphs {
+    type DataArray = number[][];
+
+    type Data = string|DataArray|google.visualization.DataTable;
+
+    interface PerSeriesOptions {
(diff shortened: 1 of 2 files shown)


danvk:
It would be nice to pull in the [tests from DT][1]. I guess that would involve making `tsc` a dev dependency?

The other issue is that these are the typings for dygraphs 1.1.1, not 2.0. I can update the typings in a follow-up.

[1]: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/c0e5ded13cb21b13e34db877a2cb6b3a9e64762e/dygraphs/dygraphs-tests.ts

mirabilos:
At this point in time, I think I’d rather keep the typings separately, both to avoid the dependency and because I cannot maintain them as I don’t know ts nor use it. (Actually I’d rather move the source base to ES5 and even get rid of babel.)

PR #859, opened 2017-05-02
Fix two inconsistencies in dygraph-externs.js

drewinglis:
1. The second argument of getOption should be marked as optional per
jsdoc at: http://dygraphs.com/jsdoc/symbols/Dygraph.html#getOption

2. 'boolean' is typo'd as 'Boolean' in updateOptions.


danvk:
Thanks. I'd trust the TypeScript definitions far more than the Closure Compiler ones.
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/dygraphs/index.d.ts

PR #862, opened 2017-05-15
fix not correct length for array

RahavLussato:
it more secure to check the length of Object.keys() .
on my case it did undefined error because the length was grater then the actual after filtering.

### DIFF
--- a/src/datahandler/bars-custom.js
+++ b/src/datahandler/bars-custom.js
@@ -29,26 +29,28 @@ CustomBarsHandler.prototype.extractSeries = function(rawData, i, options) {
   var series = [];
   var x, y, point;
   var logScale = options.get('logscale');
-  for ( var j = 0; j < Object.keys(rawData).length; j++) {
-    x = rawData[j][0];
-    point = rawData[j][i];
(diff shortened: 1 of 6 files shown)


danvk:
This doesn't make sense to me. `length` is the usual way to get the length of a JavaScript array:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/length

What makes `Object.keys` more secure?

pshevtsov:
@RahavLussato could you please provide a MWE to demonstrate and reproduce the issue you're trying to fix?

RahavLussato:
sorry i did it to specific to my case
take this example:
```
let a = [];
a[1] = 1;
a[1000] = 1;
```
the length is 1001 but actually we have  2 elements inside the array.
i can fix it by iterate over Object.keys or add an undefined check on every place we work with the array.
@danvk if you think this is a case we should handle tell me which way you prefer and i'll do it,
and if not i'll just use my fork :)

danvk:
I'd prefer not to specially handle arrays with holes. If you're trying to do a scatter plot, there are other ways.

PR #877, opened 2017-08-15
Added pixelRatio option to override canvas upscaling. Resolves #876, test included.

### LINKED ISSUE #876: Performance - option to override the canvas pixel ratio to avoid large slower graph updates
Latest version, working setup, not a bug, simple feature request, a jsfiddle would take some setup, but let me know if one's needed.

**Context:** Great library, especially for performance. I was previously using Echarts to plot about 30k data points that got updated multiple times per second, and this library, with not too much optimization, was 3-4x faster on desktop!

**Problem:** Unfortunately, it was quite bit slower than Echarts on mobile, because the resolution of the canvas matches the actual resolution of the phone's screen, instead of the CSS pixels, resulting in an enormous canvas size. It does look nice, and seriously, kudos for doing it, but it would be nice to have the user be able to turn it off for performance reasons on mobile devices.

**Proposal:** I think there should be a new option, perhaps "canvasPixelRatio", which takes a float, or a function(originalRatio): float, that can be checked in the few places we use getContextPixelRatio() to return the ratio for the canvas's display. It's nice to have a function so perhaps one can use Math.min() to limit it to 2, or something like that.

I'm happy to make a pull request for this, just let me know if it would be something we're OK with, or if there are any better ideas or suggestions. Thanks!

jeffvandyke:
Let me know about any desired changes or adjustments. I hand tested the range-selector gallery example with `pixelRatio: 3`, and it seems to behave the same and properly force scaling; the implementation followed a definitive pattern of overriding utils.getContextPixelRatio()

### DIFF
--- a/src/dygraph-options-reference.js
+++ b/src/dygraph-options-reference.js
@@ -299,6 +299,12 @@ OPTIONS_REFERENCE =  // <JSON>
     "type": "integer",
     "description": "Width, in pixels, of the chart. If the container div has been explicitly sized, this will be ignored."
   },
+  "pixelRatio": {
+    "default": "(devicePixelRatio / context.backingStoreRatio)",
+    "labels": ["Overall display"],
+    "type": "integer",
+    "description": "Overrides the pixel ratio scaling factor for the canvas's 3d context. Ordinarily, this is set to the devicePixelRatio / (context.backingStoreRatio || 1), so on mobile devices, where the devicePixelRatio can be somewhere around 3, performance can be improved by overriding this value to something less precise, like 1."
+  },
   "interactionModel": {
     "default": "...",
     "labels": ["Interactive Elements"],
--- a/src/dygraph-options-reference.js
+++ b/src/dygraph-options-reference.js
@@ -302,8 +302,8 @@ OPTIONS_REFERENCE =  // <JSON>
   "pixelRatio": {
     "default": "(devicePixelRatio / context.backingStoreRatio)",
     "labels": ["Overall display"],
-    "type": "integer",
-    "description": "Overrides the pixel ratio scaling factor for the canvas's 3d context. Ordinarily, this is set to the devicePixelRatio / (context.backingStoreRatio || 1), so on mobile devices, where the devicePixelRatio can be somewhere around 3, performance can be improved by overriding this value to something less precise, like 1."
+    "type": "float",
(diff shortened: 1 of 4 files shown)


jeffvandyke:
Just checking, any issues with merging this pull request?

src/dygraph-options-reference.js
```diff
@@ -299,6 +299,12 @@ OPTIONS_REFERENCE =  // <JSON>
     "type": "integer",
     "description": "Width, in pixels, of the chart. If the container div has been explicitly sized, this will be ignored."
   },
+  "pixelRatio": {
+    "default": "(devicePixelRatio / context.backingStoreRatio)",
+    "labels": ["Overall display"],
+    "type": "integer",
```
danvk: Does this have to be an integer? Can it be a float?

src/dygraph-options-reference.js
```diff
@@ -299,6 +299,12 @@ OPTIONS_REFERENCE =  // <JSON>
...
+    "description": "Overrides the pixel ratio scaling factor for the canvas's 3d context. Ordinarily, this is set to the devicePixelRatio / (context.backingStoreRatio || 1), so on mobile devices, where the devicePixelRatio can be somewhere around 3, performance can be improved by overriding this value to something less precise, like 1."
```
danvk: 2d context?

danvk: add a caveat here, something along the lines of "performance can be improved at the expense of resolution"

src/dygraph.js
```diff
@@ -861,7 +861,11 @@ Dygraph.prototype.resizeElements_ = function() {
   this.graphDiv.style.width = this.width_ + "px";
   this.graphDiv.style.height = this.height_ + "px";
 
-  var canvasScale = utils.getContextPixelRatio(this.canvas_ctx_);
+  var pixelRatioOption = this.getNumericOption('pixelRatio')
+
+  var canvasScale = pixelRatioOption > 0
```
danvk: Any reason not to do `pixelRatioOption || utils.getContextPixelRatio(this.canvas_ctx_)`?

src/dygraph-options-reference.js
jeffvandyke: Yep, good idea, since devicePixelRatio can often be a non-whole number for different zoom levels and certain phones.

src/dygraph.js
jeffvandyke: Probably trying to make my linter happier with 80 char lines, I switched it back here and in range-selector.js, it does read more naturally.

src/dygraph-options-reference.js
jeffvandyke: Good note.

jeffvandyke:
Hi, just wondering if you can merge in this pull request? I couldn't find any other changes asked for.

danvk:
Thanks!

PR #883, opened 2017-09-24
use whole-pixel coordinates for RRdTool-style plots, fixes #882

### LINKED ISSUE #882: option to disable subpixel dithering of lines
On real-world graphs, which are not so smooth as as the gallery demos, line dithering makes the picture not clear.

For example: https://imgur.com/a/ZsrRk

The upper image is ```dygraph```, the lower is ```rrdtool```. Despite the lesser space, ```rrdtool```'s image is much cleaner, ```dygraph```s looks like a jpeg

It would be nice to have an ```dygraph``` option to make ```rrdtool```-like graph, without introducing new colors for pixels which are only partially covered by graph lines.

volth:
Please read the guide to making dygraphs changes:
http://dygraphs.com/changes.html

Pull Requests will only be accepted if:

- You clearly explain what you're adding and why you believe it's an
  improvement. For example: "Fixes issue #123".
- You adhere to the style of the rest of the dygraphs code base.
- You write an `auto_test` for the code that you're adding.

Be sure to document any new options you add. Also be aware that PRs which add
options are likely to be rejected. dygraphs already has many options. If you
can fit your feature into one of those or implement it as a plugin, it will be
more likely to get merged.

### DIFF
--- a/src/dygraph-canvas.js
+++ b/src/dygraph-canvas.js
@@ -204,6 +204,12 @@ DygraphCanvasRenderer._drawSeries = function(e,
       point = arr[i];
     }
 
+    // RRDtool-like plot, all lines are strictly horizontal and vertical so avoid using subpixel coordinates
+    if (stepPlot && strokeWidth == 1) {
+      point.canvasx = Math.round(point.canvasx+0.5)-0.5;
+      point.canvasy = Math.round(point.canvasy+0.5)-0.5;
+    }
+
     // FIXME: The 'canvasy != canvasy' test here catches NaN values but the test
     // doesn't catch Infinity values. Could change this to
     // !isFinite(point.canvasy), but I assume it avoids isNaN for performance?


src/dygraph-canvas.js
```diff
@@ -204,6 +204,12 @@ DygraphCanvasRenderer._drawSeries = function(e,
       point = arr[i];
     }
 
+    // RRDtool-like plot, all lines are strictly horizontal and vertical so avoid using subpixel coordinates
+    if (stepPlot && strokeWidth == 1) {
+      point.canvasx = Math.round(point.canvasx+0.5)-0.5;
```
danvk: Thanks for the PR! My one concern here is that you're mutating the `point` object, which is surprising behavior for a method called `_drawSeries`. I'd feel more comfortable if you created local `canvasx` and `canvasy` variables and modified those.

Does this visual optimization also make sense for thicker strokeWidth?

volth: I recently realized that my graphs, although being visually similar to ```stepPlot``` are very different in the nature of underlying data. The data is not values which are valid from the time of introduction toward the next value or the end of the times (e.g. a central bank-imposed interest rate) but rather are differentials of a counter value (e.g. traffic counter) taken at the time. It affect almost all aspect of how those plots to be drawn (the line go to past, not to future), stacked and how the gaps in data are filled in the graphs. So it seems to end up in intruducing another graph type in additional to existing ```stepPlot```.

PR #896, opened 2017-12-11
Large speed up and benchmark for graphs with many points

vii:
Adding fields to a structure after it is created can be slow on some
browsers, like Chrome.

Adding a demo to benchmark many points, this seems to speed up my
Chromium from about 2.8-3s to about 1.8s.

Thanks to Christopher Palmer (@thecav) for the analysis and specific
suggestion.


vii:
@danvk would love some feedback! @saboya has some improvements upon the optimizations

danvk:
Thanks for this change @vii. I can confirm that it is also a speedup for me in Chrome 62.

I don't think a performance benchmark is appropriate for the gallery, so I'm going to revert that change. Have you seen `tests/dygraph-many-points-benchmark.html`? It does something similar.

danvk:
I'll hold off on cutting a new release until @saboya sends a PR.

saboya:
Hello.

The link that I made on my commit was actually just a reference, it's not an improvement over @vii 's code, it's just an implementation using a custom dataHandler on my fork of react-dygraphs. Since dygraphs releases usually take some time, I just did that in order to benefit from it right now.

Thanks for your work on dygraphs, and happy new years everyone :)

Issue #781, opened 2016-09-29
invalid charts rendering when value is constant and yRangePad = 1


ktsaou:
Hi,

Using the latest git version, when there is a chart with a **constant non-zero value**, the chart is not rendered properly.

The following images come from the same browser, same application, but different dygraph versions.

On the left the git version f6ec7be9627c2dc74776a027349c58f956a99621 and on the right the latest release:

**Note that the git version does not show the y-axis labels at all. It is like the chart is rendered below the visible area.**

![image](https://cloud.githubusercontent.com/assets/2662304/18971560/1cb55c4c-869e-11e6-850a-ccf869beb411.png)

![image](https://cloud.githubusercontent.com/assets/2662304/18971590/394d7f7e-869e-11e6-8c0a-fe9f9fb52469.png)

![image](https://cloud.githubusercontent.com/assets/2662304/18971658/7c0926f6-869e-11e6-9f8a-81bf148ff804.png)

danvk:
I merged in a major change yesterday which requires you to source a CSS file to get a proper rendering. Maybe you're running into that?
https://github.com/danvk/dygraphs/pull/678

ktsaou:
I am reading now (I see I use the killed options). I'll use the new ones and come back.
Keep in mind I made the same test 2 weeks ago, and had the same result.

danvk:
OK. If you do still see the issue, please provide a self-contained repro. There are _many_ charts in `tests/` which do not appear broken.

ktsaou:
Yeap, I included `dygraph.css`, the problem remains. I will try to make a test for it.

ktsaou:
Found it.

The problem is `yRangePad: 1`.
If I remove this, it works.

Here is a test:

```html
<!DOCTYPE html>
<html>
  <head>
    <link rel="stylesheet" href="../css/dygraph.css">
    <title>dygraph</title>
    For production (minified) code, use:
    <script type="text/javascript" src="../dist/dygraph-combined.js"></script>
    <script type="text/javascript" src="../src/extras/smooth-plotter.js"></script>
    <!--
    <script type="text/javascript" src="../dist/dygraph.js"></script>
    -->

  </head>
  <body>
    <p>Minimal example of a dygraph chart:</p>
    <div id="graphdiv"></div>
    <script type="text/javascript">
      g = new Dygraph(document.getElementById("graphdiv"),
                      "Date,Temperature\n" +
                      "2008-05-07,10\n" +
[...]

PR #909, opened 2018-02-10
fixes issue #781

ktsaou:
As described in #781, the library makes wrong decisions when the y-range is zero.
In such cases (`y0 === y1`), this PR changes:

```js
y0 -= 0.5
y1 += 0.5
```

so that the chart is properly rendered.

### DIFF
--- a/src/dygraph.js
+++ b/src/dygraph.js
@@ -2539,8 +2539,13 @@ Dygraph.prototype.computeYAxisRanges_ = function(extremes) {
 
       // special case #781: if we have no sense of scale, center on the sole value.
       if (y0 === y1) {
-          y0 -= 0.5;
-          y1 += 0.5;
+        if(y0 === 0) {
+          y1 = 1;
+        } else {
+          var delta = Math.abs(y0 / 10);
+          y0 -= delta;
+          y1 += delta;
+        }
       }
 
       if (logscale) {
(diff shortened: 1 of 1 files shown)


src/dygraph.js
```diff
@@ -2533,16 +2533,22 @@ Dygraph.prototype.computeYAxisRanges_ = function(extremes) {
     if (!ypadCompat) {
       // When using yRangePad, adjust the upper/lower bounds to add
       // padding unless the user has zoomed/panned the Y axis range.
+
+      y0 = axis.computedValueRange[0];
+      y1 = axis.computedValueRange[1];
+
+      // special case #781: if we have no sense of scale, center on the sole value.
+      if (y0 === y1) {
+          y0 -= 0.5;
```
danvk: For a log scale, this could drop y below zero, which will produce `NaN`s. In the `logscale` branch of the conditional on line 2546, you can halve y0 and double y1 instead.

ktsaou:
There are 3 cases where `y0 === y1`:

1. they are both zero
2. they are both positive
3. they are both negative

I think this will properly handle all cases, for both normal and logscale rendering:

```js
      if (y0 === y1) {
      	if(y0 === 0) {
      	  y1 = 1;
      	} else {
      	  var delta = Math.abs(y0 / 2);
          y0 -= delta;
          y1 += delta;
      	}
      }
```

So, only when it is zero, set `y1 = 1` (from the charts I render, I see this must be a default for other cases too - for example when stacked charts have only zeros, the library already visualizes 0 to 1).

Otherwise provide a range of -1/2 to +1/2 of what already the number is. So, do not change the sign. If it is negative keep it negative, if it is positive keep it positive.

Do you agree?

Of course I agree there should be a unit test for this part of the code - its absence is what caused the bug in the first place. However, I need some guidance - I don't have a clue how your unit testing works, or how to configure one.

PR #935, opened 2018-06-14
Allow setSelection to trigger highlightCallback and do so in synchronizer

lightvector:
Currently, there appears to be either a bug or a surprising behavior with extras/synchronizer.js where the synchronizer's selection of a data series in the synchronized graph fails to trigger the user's highlightCallback function.

This pull request adds a parameter to setSelection that allows specifying triggering of highlightCallback, which is then used by synchronizer.js to trigger the callback.

All existing tests from "npm run test" appear to still pass.


ikmolbo:
Excellent. I came here with the same problem, which you have resolved with this PR. Thanks.

PR #936, opened 2018-06-19
Fix drawAxes to allow one of two Y axes to be drawn

DJHig:
Hello. This fix is for a bug I noticed that prevented the drawAxes function to properly show only one Y axis when there is data for both. I added a test to cover Y1 drawn, but not Y2 and one for Y2 drawn, but not Y1. 

Separately, the docs state the default is that Y2 is not drawn, but the code has all axes drawn by default. I am not sure which is correct so I did not change either.

Thanks for the great work!

### DIFF
--- a/auto_tests/tests/css.js
+++ b/auto_tests/tests/css.js
@@ -53,6 +53,7 @@ it('testExplicitParamSize', function() {
 
 // Verifies that setting a style on the div works.
 it('testExplicitStyleSize', function() {
+  this.timeout(5000);
   var opts = {
   };
   var graph = document.getElementById("graph");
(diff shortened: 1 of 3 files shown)


DJHig:
I fixed the broken tests from the first push by refactoring some code.
Failing now due to timeout in a random test. Not sure if possible to fix that.

mirabilos:
I have merged this without the test timeout increase, as it was not needed. Thank you!

mirabilos:
This unfortunately caused a regression: https://dygraphs.com/tests/unboxed-spark.html disables drawing of the lines, and this did not work any more correctly.

fixed by https://github.com/danvk/dygraphs/commit/046bdec9ecbdf475f5283fa41efde8f2bf13b1d4

PR #946, opened 2018-10-28
Adds offset config for Legend div in 'follow' mode

zinahe:
. At the moment, legend offset in **follow** mode is hard-coded at 50px
. Adds new config options **topLegendOffset** and **leftLegendOffset** in src/dygraph-options-reference.js
. Implements legend offsets only when (legendMode === 'follow') in src/plugins/legend.js

Please read the guide to making dygraphs changes:
http://dygraphs.com/changes.html

Pull Requests will only be accepted if:

- You clearly explain what you're adding and why you believe it's an
  improvement. For example: "Fixes issue #123".
- You adhere to the style of the rest of the dygraphs code base.
- You write an `auto_test` for the code that you're adding.

Be sure to document any new options you add. Also be aware that PRs which add
options are likely to be rejected. dygraphs already has many options. If you
can fit your feature into one of those or implement it as a plugin, it will be
more likely to get merged.

### DIFF
--- a/src/dygraph-options-reference.js
+++ b/src/dygraph-options-reference.js
@@ -821,7 +821,20 @@ OPTIONS_REFERENCE =  // <JSON>
     "labels": ["Data"],
     "type": "Dygraph.DataHandler",
     "description": "Custom DataHandler. This is an advanced customization. See http://bit.ly/151E7Aq."
+  },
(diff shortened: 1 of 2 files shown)


mirabilos:
I have manually merged this.

PR #952, opened 2019-01-30
Using var instead let in order to support old browsers

matheussouza9:
My applications needs to run on Firefox 43 (unfortunaly I can't update it) and it doesn't support `let` identifier. So, this simple pull request changes it to `var`.

### DIFF
--- a/src/dygraph.js
+++ b/src/dygraph.js
@@ -2788,7 +2788,7 @@ function validateNativeFormat(data) {
   if (typeof firstX !== 'number' && !utils.isDateLike(firstX)) {
     throw new Error(`Expected number or date but got ${typeof firstX}: ${firstX}.`);
   }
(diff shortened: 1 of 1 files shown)


danvk:
How are you sourcing dygraphs? The distribution version should use var.

matheussouza9:
Dygraphs installed from npm. I'm using with Angular 7.
`import Dygraph from 'dygraphs'`
Importing it like this, it loads `dygraph.js` from **`src`** instead `dygraph.min.js` from **`dist`**.
To solve the problem, I've added the path to the minified file in `angular.json` scripts section like
`"scripts": [
              "node_modules/dygraphs/dist/dygraph.min.js"
            ]` 
and "importing" (only declaring) in .ts files like this:
`declare let Dygraph: any;`

danvk:
There's also an `index.es5.js` file that you could try importing instead.

Issue #611, opened 2015-05-21
Demo broken: Drawing zoom option errors


lBowlin:
The drawing demo is now broken. The change introduced to move the mouseup and mousemove to the document most likely broke this (see https://github.com/danvk/dygraphs/pull/472). Mousemove and mousup are now undefined: 

Dygraph.defaultInteractionModel.mousemove(event, g, context);
Dygraph.defaultInteractionModel.mouseout(event, g, context);
Dygraph.defaultInteractionModel.mouseup(event, g, context);

lBowlin:
I noticed in 1.1.1 this no longer issues an undefined error, the references to mousemove and mouseup were removed in favor of just referencing the default mousedown. 

This should work, but when zooming the mouseup event triggers, but context.isZooming is always false which results in never completing the zoom function.

I've been looking into it, but can't figure out where isZooming becomes false. 

This doesn't just break the demo, but any other custom interaction model implementation that wishes to make use of the defaults.

pm31:
I want to define an interaction model as defined <a href='http://www.technogumbo.com/2013/04/Dygraphs-Tablet-and-Phone-Custom-Interaction-Touch-Point-Selection/'>here</a> : 
<pre><code>interactionModel:
{
mousedown: Dygraph.defaultInteractionModel.mousedown,
mousemove: Dygraph.defaultInteractionModel.mousemove,
mouseup: Dygraph.defaultInteractionModel.mouseup,
touchstart: newDygraphTouchstart,
touchend: Dygraph.defaultInteractionModel.touchend,
touchmove: Dygraph.defaultInteractionModel.touchmove
} </code><pre>

I m having the same problem mousemove and mousup are undefined. My purpose is to set a different behaviour when the graph is displayed on touchscreen. Is there a different way to do that ?

HenkV74:
Is there a solution available? If I include only the mousedown it doesn't work correctly. 

Basically what I want to do is using the defaultmodel and when the user makes a selection ( just like he does when zooming ), it should check the state of the ctrl-key. If not pressed, go on with the default zooming, otherwise do my stuff and finish the event.

wookhuk:
A little late to the convo but @lBowlin a quick workaround is to destroy the graph and recreate it without the custom interactionModels. Not ideal but it does get the job done. I'll try to see if I can address this issue.

bbirand:
The drawing demo somehow works now. But when I try to replicate in my own project, I get the same errors as above:
```
[...]

PR #953, opened 2019-03-17
Fix issue #611 in the drawing demo gallery

mp035:
This pull request fixes the broken zoom functionality  and console errors in the drawing demo.  It simply adds the willDestroyContextMyself attribute, and removes the now unimplemented functions in the defaultInteractionModel.  Comments have also been added to explain to users why calls are missing.

Sorry about the double commit, still getting a handle on pull request functionality.  13887ef corrects a problem where synchronizer.js would block the users drawCallback when updating the zoom of connected charts.

### DIFF
--- a/gallery/drawing.js
+++ b/gallery/drawing.js
@@ -119,6 +119,9 @@ Gallery.register(
             valueRange: valueRange,
             labels: [ 'Date', 'Value' ],
             interactionModel: {
+              // the next line is required when using the 
+              // defaultInteractionModel mousedown function.
+              willDestroyContextMyself: true, 
               mousedown: function (event, g, context) {
                 if (tool == 'zoom') {
                   Dygraph.defaultInteractionModel.mousedown(event, g, context);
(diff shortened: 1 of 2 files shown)


mirabilos:
Is this (and #611 therefore) still pertinent? It doesn’t throw any console warnings for me…

mirabilos (dismissed):
is the bug still there?

mp035:
Hi, yep, the bug is still there.  Select the zoom functionality in [https://dygraphs.com/gallery/#g/drawing](https://dygraphs.com/gallery/#g/drawing) with the console open.

mirabilos:
@mp035 thank you for both retesting and the reproducer. I *think* I get what you’re doing here now.

The [unrelated synchroniser change](https://github.com/danvk/dygraphs/pull/953/commits/13887ef95bb2c28849d56f8e59ddfbc0c55cd23e) however, do you also have a comment/docs/reproducer/background info on that? I’ve no problem just merging this PR all in one, but I want to make sure I don’t break anything that works, at this point… (Debian is almost frozen, and I want to make a v2.2.1 with all those tiny fixes that arose since v2.2.0)

mp035:
@mirabilos , sorry, I missed your comment.  It has been nearly 4 years since I worked on this so, I am a little fuzzy on the details.  I think the issue was that the end-user could not rely on their callback being called on every redraw because sometimes synchronizer.js would block for its own purposes and never propagate to the next callback.   I see you have merged it anyway, I just didn't want to leave you hanging.

mirabilos:
Mark Pointing dixit:

>@mirabilos , sorry, I missed your comment.  It has been nearly 4 years
>@since I worked on this so, I am a little fuzzy on the details. I think

Don’t worry, I understand.

>@the issue was that the end-user could not rely on their callback being
>@called on every redraw because sometimes synchronizer.js would block
>@for its own purposes and never propagate to the next callback. I see

I also think it fixed another bug, crosshairs<->synchroniser interop.

>@you have merged it anyway, I just didn't want to leave you hanging.

Yes, thank you!

bye,
//mirabilos
-- 
Solange man keine schmutzigen Tricks macht, und ich meine *wirklich*
schmutzige Tricks, wie bei einer doppelt verketteten Liste beide
Pointer XORen und in nur einem Word speichern, funktioniert Boehm ganz
hervorragend.		-- Andreas Bogk über boehm-gc in d.a.s.r

mp035:
Thanks for the update, perhaps that's why I didn't pull [13887ef](https://github.com/danvk/dygraphs/pull/953/commits/13887ef95bb2c28849d56f8e59ddfbc0c55cd23e) into my personal codebase.  There was a massive amount of cursor customization in my application so I may not have needed it.

PR #955, opened 2019-03-26
Added strokeStyle option to crosshair plugin to allow color choice.

rishabhkohli:
I needed this functionality for my graphs because I am working in a dark theme. The default color was hardly visible.

Maybe the option should be called 'color' instead of strokeStyle, or maybe color can be an object inside strokeStyle
But I let it remain consistent with the current variables for now.

What do you think?

### DIFF
--- a/src/extras/crosshair.js
+++ b/src/extras/crosshair.js
@@ -19,6 +19,7 @@ Dygraph.Plugins.Crosshair = (function() {
     this.canvas_ = document.createElement("canvas");
     opt_options = opt_options || {};
     this.direction_ = opt_options.direction || null;
+    this.strokeStyle = opt_options.strokeStyle || "rgb(0,0,0,0.3)";
   };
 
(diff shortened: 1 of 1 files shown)


src/extras/crosshair.js
```diff
@@ -19,6 +19,7 @@ Dygraph.Plugins.Crosshair = (function() {
     this.canvas_ = document.createElement("canvas");
     opt_options = opt_options || {};
     this.direction_ = opt_options.direction || null;
+    this.strokeStyle = opt_options.strokeStyle || "rgb(0,0,0,0.3)";
```
mirabilos, line 22: `rgba`

mirabilos: `"rgba(0, 0, 0, 0.3)"` to be exact

mirabilos (changes requested):

PR #964, opened 2019-05-23
Add missing semicolon in externs

jkuszmaul:


danvk:
Thanks. Are you using dygraphs with Closure Compiler?

jkuszmaul:
I am not compiling dygraph myself; I'm taking the released dygraph.min.js you provide and using the externs file with closure to get the declarations. 

I should probably make a pull request for them as well, but I also had to:

- Add an `@externs` tag in the file comment for the externs file.
- Changed the `HTMLDivElement` to just an `Element`, because I was too lazy to find the proper way to get an `HTMLDivElement` object out of the DOM (I'm not a javascript person normally--I just needed to create a small plotting webpage for a project).

PR #1012, opened 2021-08-30
There's 365.2425 days in a gregorian year

atg:
I was [idly searching github](https://github.com/search?q=365.2524&type=code) for bugs and found that many references to the constant 365.2524 originate from this library. The actual constant is [365.2**4**2**5**](https://en.wikipedia.org/wiki/Gregorian_calendar). It doesn't really matter at all, but I thought it would be nice to correct it.


przmv (approved):
LGTM

danvk:
Amazing! The truth will eventually come out? Nice to know that dygraphs has had influence, even if it's been spreading misinformation :)

PR #1013, opened 2021-10-21
Calculate legend positioning relative to closest data point

acontreras89:
This is done by checking the current highlighted series.

### DIFF
--- a/src/plugins/legend.js
+++ b/src/plugins/legend.js
@@ -107,12 +107,18 @@ Legend.prototype.select = function(e) {
     var area = e.dygraph.plotter_.area;
     var labelsDivWidth = this.legend_div_.offsetWidth;
     var yAxisLabelWidth = e.dygraph.getOptionForAxis('axisLabelWidth', 'y');
+    // find the closest data point by checking the currently highlighted series,
+    // or fall back to using the first data point available
+    var highlightSeries = e.dygraph.getHighlightSeries()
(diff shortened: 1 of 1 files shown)


acontreras89:
Hey @mirabilos, thanks for merging our PR! 🎉 

After some time using our own version of the library (which included this change), we spotted a little bug 🐛 eventually `point` would be undefined, causing the following error:

plot-zoom-out-of-sync-error

Not sure why this was happening,* we fixed it by simply adding a guard in commit https://github.com/redradix/dygraphs/commit/1a74f4d0ed24c75ed5ba1331496ec041d6695159.

_* It is possible that when the user is moving the mouse, we try to calculate the next legend position before the highlight series has been updated. In this case, a point for the highlighted series may not be available 🤷‍♂️_

mirabilos:
Hi @acontreras89, interesting. Why not simply default to `points[0]` then? Is there any negative effect to expect from returning early, or will the code just be called again once points are available again?

acontreras89:
@mirabilos falling back to `points[0]` would most likely be a better idea.

Our use case is very sensitive in terms of performance, so we opted for the guard assuming this was an unnecessary legend computation (following on my previous comment.)

But honestly, I think we simply did not put that much effort into figuring out what was really happening. I just wanted to let you know this can happen now that the PR (finally) got merged 🙃

Issue #571, opened 2015-03-29
labelsKM{B,G2} behave oddly for values in (0, 1)


danvk:
`labelsKMB` has no effect on values between 0 and 1. You might expect it to use "m", "µ" and "n" lowercase prefixes.

`labelsKMG2` does something bizarre. In `testSmallLabelKMG2` I see these labels:

```
['0', '0.48u', '0.95u', '1.43u', '1.91u']
```

These aren't even multiples of powers of two. And it's unclear to me whether those prefixes are used for binary units (e.g. "u" to mean 2^-20).

mirabilos:
Working on this (see #994 comments), but this is really weird, there must be some kind of data smoothing going on?

With my current WIP code, these are just printed as (decimal, supposedly!) scientific (and, oh, why does this use the lowercase `e` for this, this reads so weird), and Travis-CI logs:

       [
         "0"
         "4.77e-7"
         "9.54e-7"
         "1.43e-6"
         "1.91e-6"
       ]

The median is supposed to be the second value IIUC.

    .0000009540 // out
    .0000010000 // in

The delta is really small, but it is there…

mirabilos:
It’s a bug in `numericTicks()`.

mirabilos:
tbh I’m not convinced in “small 2-based” units for the axes at all, after all 0.5, 0.25, etc. are the corresponding powers of two

mirabilos:
Hm, or maybe it’s _not_ a bug in `numericTicks()` and the input values should just be the expected ones (e.g. `2^⁻ˣ`)…

mirabilos:
Yeah, I guess it’s just the test.

PR #1022, opened 2022-09-10
fix label units

### LINKED ISSUE #994: Extend labelsKMB for milli, micro, nano
The option "labelsKMB" in dygraphs 2.1.0 allows automatic labeling with kilo, mega and giga prefixes.
It would be great if that could be extended to milli (m), micro (&#181;) and nano (n).

Or is there maybe another option that allows this already?

By the way: the right unit prefix for kilo would be a lower case k (https://en.wikipedia.org/wiki/Unit_prefix).

mirabilos:
This fixes #994 and #571 by:

- using the correct SI units for `labelsKMB`
- using the correct IEC 60027-2 units for `labelsKMG2`
  - unless `labelsKMB` is _also_ given, in which case the “legacy” units (basically the same as `labelsKMB` except with `K` instead of `k` for Kibi) are used instead
  - for the nōn-KMB KMG2 case, use binary float-style “`p-10`” notation for the “small” units; the way `numericTicks()` operates makes it basically necessary to have “small” units for all labelling styles
- fix the KMG2-related small label tests by picking values that make checking the labels properly possible

### DIFF
--- a/auto_tests/tests/axis_labels.js
+++ b/auto_tests/tests/axis_labels.js
@@ -646,11 +646,11 @@ it('testSmallLabelKMB', function() {
                    Util.getYLabels());
 });
 
-it('testSmallLabelKMG2', function() {
+it('testSmallLabelKMG2legacy', function() {
   var data = [];
   data.push([0, 0]);
-  data.push([1, 1e-6]);
-  data.push([2, 2e-6]);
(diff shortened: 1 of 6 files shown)


mirabilos:
Hmm, I should probably extend the gallery/tests for the third possibility as well…

PR #1035, opened 2023-04-09
Synchronizer plugin - selectClosest option

n-gist:
Adds `selectClosest` option to synchronizer plugin so it selects values closest to desired. Defaults to false.
It can be helpful when graphs have not identical x axes data. (For example, when a graph has only one of each n values of another, or the values are different, but are close visually)

As the added search method `closestIdx()` is a bit slower (about 3 time slower than `getRowForX()`), I have added some optimization logic:
If the option is turned on, it checks graphs rows lengths first. If they are equal, it assumes that axes have identical values and thus tries to use `getRowForX()`. Otherwise, or if the value was not found, it uses new `closestIdx()` search function (uses binary search algorithm).

### DIFF
--- a/src/extras/synchronizer.js
+++ b/src/extras/synchronizer.js
@@ -36,6 +36,11 @@
  *
  * You may also set `range: false` if you wish to only sync the x-axis.
  * The `range` option has no effect unless `zoom` is true (the default).
+ *
+ * Synchronizer selection option selects points by exact values on x-axes.
+ * If graphs have different x-axes values, you may use
+ * `selectionClosest: true` to select points closest to hovered ones,
+ * corresponding to their x values.
  */
 
 /* loader wrapper to allow browser use and ES6 imports */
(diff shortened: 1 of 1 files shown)


n-gist (commented):
Switched to use layout_.points instead of graph.getValue() for closest point search. Tested on two synchronized graphs (of 1000 and 100 rows resp) with 1mil iterations, got about 50% performance improvement when unzoomed and more if zoomed in

n-gist:
Also filtered when the found point is out of graph view. This should be done for crosshair plugin too, I'll open separate pr for it

mirabilos:
n-gist dixit:

>got about 50% performance improvement

Sweet!

>Also filtered when the found point is out of graph view. This should be
>done for crosshair plugin too, I'll open separate pr for it

OK, thanks!

I don’t have the bandwidth to look at it currently (was ill for over two
weeks and have to catch up on missed things now) but I’ll look at them
eventually.

n-gist:
When searching for closest point on synced graphs, x axis values are compared. Fine for default graphs, but for those which are in `stepPlot` mode, left point should be selected, as it represents actual strict value for corresponding selection on the first graph

For example, selecting `4` on one graph, and having values on another such as `[...,2,5,...]`, `5` is selected as closest by value. Good for default mode as closest for interpolated value, but for `stepPlot` graph `2` should be selected, since it is `2` for whole `2 -> 5` period

This can be done as a separate option, but it seems logical to make it the default

n-gist:
Adding small fix for stepPlot mode graphs. If synchronizing graph has exact x axis value, return it, even if it is ending point of a step period

n-gist:
I have added a note to comment section at the top, and made code style fixes

Also created an example to test the option https://jsfiddle.net/3d80v5xc/
graph 2 has no points on `3,4,5` x values
graph 3 has no points on `5,6,7`
When you hover over these values on graph 1, nothing is selected on other two graphs
Toggle `selectionClosest` to `true`, so that it starts to select points closest to desired

PR #1038, opened 2023-05-07
Crosshair plugin - initial canvas size fix

n-gist:
Crosshair plugin creates canvas to draw its lines, but if the graph is small enough, this canvas can block interaction with other elements until you hover over a graph giving it a chance to resize. Canvas initializes with 300x150 which is default [Canvas element - Wikipedia](https://en.wikipedia.org/wiki/Canvas_element#:~:text=By%20default%2C%20both%20the%20canvas,and%20150%20screen%20pixels%20high.)

In the example below it overlaps the '+' button preventing clicking it

dgCanvas

To fix it I have set the size to [0,0] in plugin constructor.

### DIFF
--- a/src/extras/crosshair.js
+++ b/src/extras/crosshair.js
@@ -28,18 +28,18 @@ Dygraph.Plugins.Crosshair = (function _extras_crosshair_closure() {
 
   var crosshair = function crosshair(opt_options) {
     this.canvas_ = document.createElement("canvas");
-    this.updateCanvasSize(0, 0)
     opt_options = opt_options || {};
     this.direction_ = opt_options.direction || null;
     this.strokeStyle_ = opt_options.strokeStyle || "rgba(0, 0, 0, 0.3)";
   };
 
   crosshair.prototype.updateCanvasSize = function updateCanvasSize(width, height) {
-    this.canvas_.width = width
-    this.canvas_.height = height
-    this.canvas_.style.width = width.toString(10) + 'px'    // for IE
-    this.canvas_.style.height = height.toString(10) + 'px'  // for IE
-  }
+    if (width === this.canvas_.width && height === this.canvas_.height) return;
+    this.canvas_.width = width;
+    this.canvas_.height = height;
+    this.canvas_.style.width = width + 'px';    // for IE
+    this.canvas_.style.height = height + 'px';  // for IE
+  };
 
   crosshair.prototype.toString = function toString() {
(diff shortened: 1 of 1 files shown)


src/extras/crosshair.js
```diff
@@ -28,11 +28,19 @@ Dygraph.Plugins.Crosshair = (function _extras_crosshair_closure() {
...
+    this.updateCanvasSize(0, 0)
     opt_options = opt_options || {};
     this.direction_ = opt_options.direction || null;
     this.strokeStyle_ = opt_options.strokeStyle || "rgba(0, 0, 0, 0.3)";
   };
 
+  crosshair.prototype.updateCanvasSize = function updateCanvasSize(width, height) {
+    this.canvas_.width = width
+    this.canvas_.height = height
+    this.canvas_.style.width = width.toString(10) + 'px'    // for IE
```
mirabilos, line 40: 10 is default

src/extras/crosshair.js
```diff
@@ -28,11 +28,19 @@ Dygraph.Plugins.Crosshair = (function _extras_crosshair_closure() {
 
   var crosshair = function crosshair(opt_options) {
     this.canvas_ = document.createElement("canvas");
+    this.updateCanvasSize(0, 0)
```
mirabilos, line 31: would `1, 1` work as well? bit afraid of hitting corner cases here

src/extras/crosshair.js
mirabilos: and the `toString` is pretty redundant, the entire rest of the code doesn’t use it either, so unless you can provide a very good reason, please revert to the string concatenation

n-gist: Confused by eslint [radix rule](https://eslint.org/docs/latest/rules/radix). Using 10 as optimization to skip possible checks inside. No need on object that is a number for sure

src/extras/crosshair.js
n-gist: I think size is not important until canvas is added to DOM. Moved to `activate` callback, where it can be set to graph actual size

n-gist:
I have also added a size change check before updating it to remove possible unnecessary layout updates and remove a bit of garbage on mouse movements

mirabilos:
n-gist dixit:

>I think size is not important until canvas is added to DOM. Moved to
>`activate` callback, where it can be set to graph actual size

Ah, good plan.

PR #1039, opened 2023-05-16
Crosshair plugin - out of graph bounds selection filter

n-gist:
When the graph is zoomed in, crosshair can select points out of view area, drawing lines off graph.
Below is example where the area on the left is caused by `axisLabelWidth`, so it is inside crosshair drawing canvas.

chBounds

The same happens on the right side when using `rightGap`. Filtered by checking `selectedPoint.x` to be in [0,1]

### DIFF
--- a/src/extras/crosshair.js
+++ b/src/extras/crosshair.js
@@ -67,8 +67,8 @@ Dygraph.Plugins.Crosshair = (function _extras_crosshair_closure() {
     ctx.strokeStyle = this.strokeStyle_;
     ctx.beginPath();
     
-    var point = e.dygraph.selPoints_[0];
-    if (point.x >= 0 && point.x <= 1) {
+    var p_x = e.dygraph.selPoints_[0];
+    if (p_x.x >= 0 && p_x.x <= 1) {
       var canvasx = Math.floor(e.dygraph.selPoints_[0].canvasx) + 0.5; // crisper rendering
   
       if (this.direction_ === "vertical" || this.direction_ === "both") {
@@ -79,7 +79,9 @@ Dygraph.Plugins.Crosshair = (function _extras_crosshair_closure() {
 
     if (this.direction_ === "horizontal" || this.direction_ === "both") {
       for (var i = 0; i < e.dygraph.selPoints_.length; i++) {
-        var canvasy = Math.floor(e.dygraph.selPoints_[i].canvasy) + 0.5; // crisper rendering
+        var p_y = e.dygraph.selPoints_[i];
+        if (p_y.y < 0 || p_y > 1) continue;
+        var canvasy = Math.floor(p_y.canvasy) + 0.5; // crisper rendering
         ctx.moveTo(0, canvasy);
         ctx.lineTo(width, canvasy);
       }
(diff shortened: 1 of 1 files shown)


src/extras/crosshair.js
```diff
@@ -66,17 +66,22 @@ Dygraph.Plugins.Crosshair = (function _extras_crosshair_closure() {
...
+        ctx.moveTo(canvasx, 0);
+        ctx.lineTo(canvasx, height);
+      }
     }
 
     if (this.direction_ === "horizontal" || this.direction_ === "both") {
       for (var i = 0; i < e.dygraph.selPoints_.length; i++) {
-        var canvasy = Math.floor(e.dygraph.selPoints_[i].canvasy) + 0.5; // crisper rendering
+        var p_y = e.dygraph.selPoints_[i];
+        if (p_y.y < 0 || p_y > 1) continue;
```
mirabilos, line 83: I think a `.y` is missing here.

I’d appreciate if you could test that everything still works right in git master past at least commit 295b8297ed8e56923bac8e97fa32a4c9e4ca04c0.

n-gist: @mirabilos yes, `.y` was missing there, I see you have it covered

Glad to see you have merged my PRs for crosshair plugin, so I can remove my patches after releasing update.

Could you check my other two PRs for synchronizer plugin #1035 and #1037? Second one was closed by mistake imo, because I included incorrect code in the post that is not actually part of the final fix.

I'm using dygraph with all these fixes combined for over a year in my project and it is stable. I haven't found anything I'd like to fix or add to this yet, other than fixing the animation when zooming out vertically, but I haven't found the time to try to do that yet. It is not so critical

I'm not sure how to inject current version into my project for testing, it is installed through yarn with loading patched plugins from other folder (probably I could just overwrite plugins files inside /node_modules/dygraphs/ of my project, and turn off my patched ones)
But I have checked current master code for all 4 merged PRs. Looks like all code is at its place. I will find a way to actually test it if you decide whether to merge my PRs for synchronizer or not, just to not do it twice.

Thank you

mirabilos:
On Mon, 13 Jan 2025, n-gist wrote:

>@mirabilos yes, `.y` was missing there, I see you have it covered

OK, thanks.

>Glad to see you have merged my PRs for crosshair plugin, so I can
>remove my patches after releasing update.

You’re welcome.

>Could you check my other two PRs for synchronizer plugin #1035 and
>#1037? Second one was closed by mistake imo, because I included
>incorrect code in the post that is not actually part of the final fix.

I had a look but it needed more headspace to review than I had,
I’ll look again when I next work on Dygraphs. (There’s something
that needs probably a lot of work: it uses an old version of
twitter-bootstrap that’s now unsupported and the Debian maintainers
of that ask me to upgrade, but I’d rather remove the use of that
entirely and add minimal CSS and (ideally no) JS directly to the
project so the documentation/site renders… similar enough to how
it currently looks. But I’m no frontend person…

>I'm using dygraph with all these fixes combined for over a year in my
>project and it is stable. I haven't found anything I'd like to fix or
>add to this yet, other than fixing the animation when zooming out
>vertically, but I haven't found the time to try to do that yet. It is
>not so critical

Hmm, okay. I’ll comment on #1037 directly a bit, too.

>I'm not sure how to inject current version into my project for testing,
>it is installed through yarn with loading patched plugins from other

I don’t know yarn, but could you not simply do an alpha release
on your own namespace and then point packages.json or whatever
it uses to that?

>I will find a way to actually test it if you decide whether to merge my
>PRs for synchronizer or not, just to not do it twice.

Fair.

PR #1050, opened 2024-10-22
Warn if points are out of range

deribaucourt:
Some users may improperly define the range of their axes, leading to points being drawn outside of the chart area. In the worst cases, if all points are out of range, the chart will not be drawn at all.

Adding a warning can help users identify this issue and correct it.

Note that may spew a lot of warnings if their are a lot of points.

---

Such a warning helped me debug the usage of netdata. See https://github.com/netdata/netdata-cloud/issues/1048

### DIFF
--- a/src/dygraph-layout.js
+++ b/src/dygraph-layout.js
@@ -247,12 +247,14 @@ DygraphLayout.prototype._evaluateLineCharts = function() {
     var axis = this.dygraph_.axisPropertiesForSeries(setName);
     // TODO (konigsberg): use optionsForAxis instead.
     var logscale = this.dygraph_.attributes_.getForSeries("logscale", setName);
+    var outOfXBounds = 0, outOfYBounds = 0;
 
     for (var j = 0; j < points.length; j++) {
       var point = points[j];
 
       // Range from 0-1 where 0 represents left and 1 represents right.
       point.x = DygraphLayout.calcXNormal_(point.xval, this._xAxis, isLogscaleForX);
+      outOfXBounds += (point.x < 0) || (point.x > 1);
       // Range from 0-1 where 0 represents top and 1 represents bottom
       var yval = point.yval;
       if (isStacked) {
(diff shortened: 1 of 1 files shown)


mirabilos:
Let’s hope the extra if isn’t going to be a performance issue…

On Tue, 22 Oct 2024, Enguerrand de Ribaucourt wrote:

>Note that may spew a lot of warnings if their are a lot of points.

How about warning only once? Perhaps just adding ((x<0)||(x>1)) to
a running counter, so we have the comparisons’ conditionals but no
extra conditional jump in the inner loop.

bye,
//mirabilos

deribaucourt:
@mirabilos Done, thanks a lot for your review!

src/dygraph-layout.js
```diff
@@ -247,12 +247,14 @@ DygraphLayout.prototype._evaluateLineCharts = function() {
...
     // TODO (konigsberg): use optionsForAxis instead.
     var logscale = this.dygraph_.attributes_.getForSeries("logscale", setName);
+    var outOfXBounds = 0, outOfYBounds = 0;
 
     for (var j = 0; j < points.length; j++) {
       var point = points[j];
 
       // Range from 0-1 where 0 represents left and 1 represents right.
       point.x = DygraphLayout.calcXNormal_(point.xval, this._xAxis, isLogscaleForX);
+      outOfXBounds += ((point.x<0)||(point.x>1));
```
mirabilos, line 257: nitpick `(point.x < 0) || (point.x > 1)`, I had left out the spaces in the mail to convey the idea and not go over normal eMail line length, and the outer parenthesēs are redundant here

src/dygraph-layout.js
```diff
@@ -269,6 +271,14 @@ DygraphLayout.prototype._evaluateLineCharts = function() {
         }
       }
       point.y = DygraphLayout.calcYNormal_(axis, yval, logscale);
+      outOfYBounds += ((point.y<0)||(point.y>1));
```
mirabilos, line 274: same

src/dygraph-layout.js
```diff
@@ -269,6 +271,14 @@ DygraphLayout.prototype._evaluateLineCharts = function() {
...
+    }
+
+    if (outOfXBounds > 0) {
```
mirabilos, line 277: I wonder whether this even makes sense for the x axis… or rather, when zooming in, some points are always going to be out of bounds, also for the y axis. Maybe the checks should be conditional on not having zoomed in? Unless they are already excluded then…

Perhaps we need to play around a bit with this.

deribaucourt: Interesting question!
If I zoom in, the warning is only displayed for 2 points (those that allow drawing the line going out of the edge it appears). So it looks like all other points are excluded. 

The rest of the time, x axis warnings are really interesting. In my use case for instance, I needed them to debug a graph not being shown because the x axis was improperly configured.

I think we would be good if we remove 2 points from the x axis out of bounds counter. It's not going to generate false negatives except for graphs with just 2 points. While providing protection for zooming.

image

src/dygraph-layout.js
deribaucourt: Done

src/dygraph-layout.js
deribaucourt: Done

PR #1057, opened 2025-03-13
Fix event listener violation for touchstart, touchmove event

n-gist:
This fixes violation reports in console (at Verbose level)
Happens on graph creation, can be checked on https://dygraphs.com/

Screenshot_1


mirabilos:
This seems to lead to `clickCallback` not being called any more.

Unfortunately, not all browsers support the form of `addEventListener` with an options object; in particular, PhantomJS doesn’t.

Could you please come up with a way to detect whether it is supported?

PR #1058, opened 2025-03-15
EventListeners violation - more selective options

n-gist:
@mirabilos I'm wondering, if PhantomJS supports `touchstart` and `touchmove` at all, so, if it discards them, but wants boolean for everything else, maybe we just pass options object only for touch events, and `false` fot others, like it was before
Can you check this?

### DIFF
--- a/src/dygraph-utils.js
+++ b/src/dygraph-utils.js
@@ -101,11 +101,10 @@ export var getContext = function(canvas) {
  * EventListener options
  */
 function _eventListenerOptions(type) {
-  const options = { capture: false };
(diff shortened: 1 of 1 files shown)


mirabilos:
Great idea! This worked.

----

Oh, oops, I only now on the web see you entered this as separate pull request instead of just a reply. But in the end, I wrote pretty much the same code.

n-gist:
Yeah sorry, it was first time when I updated the branch of merged PR, expected new commit should appear, but it didn't, so I just created new one

